PR #23, opened 2025-07-08
22 read auth token from http only cookie

### LINKED ISSUE #22: read auth token from http only cookie
## Description

modify auth part of endpoints to read http only cookies as well as from body

arturasmckwcz:
closes #22


need4deed (approved):
zzzcode.ai approved

PR #142, opened 2025-11-07
Prepare Docker image for deployment

drndivoje:
- use type orm migrations and disable synchronization 
- bootstrap container to populate database before startup (contains all data used fro seed functions)
- fix docker compose local setup
- add public certificate need for AWS deployment


src/data/entity/event/event_translation.entity.ts
```diff
@@ -53,7 +53,10 @@ export default class EventTranslation {
   @Column({ type: "varchar", length: 256, nullable: true })
   followupText?: string;
 
-  @ManyToOne(() => EventN4D, (eventn4d) => eventn4d.eventTranslation)
+  @ManyToOne(() => EventN4D, (eventn4d) => eventn4d.eventTranslation, {
+    onDelete: 'CASCADE',
+    createForeignKeyConstraints: false
```
arturasmckwcz: just curious:
what problem is that causing?

arturasmckwcz (approved):

PR #146, opened 2025-11-19
145 separate ddl and data seeding in bootstrap

### LINKED ISSUE #145: separate DDL and data seeding in bootstrap
## Description

currently bootstrap loads a single dump file containing schema and data. data is going to be updated right before going production so it would be good to decouple it from schema.

arturasmckwcz:
closes #145


arturasmckwcz:
> LGTM! As potential improvement on this I would have schema as a part of TypeORM migration file and only have data.sql as a part of bootstrap container

but schema has to be loaded first, before data
or maybe both of them could go as migrations then bootstrap would be redundant WDYT?

PR #287, opened 2026-02-13
Extend opportunity-id-mock

HansKre:
## Description

For https://github.com/need4deed-org/fe/pull/227


arturasmckwcz:
> @arturasmckwcz what do you think of using an Enum with keys for skills & activities, which fe would use to display the localized string?

@HansKre those might get updated and then we'd need refactoring of the code instead manipulating data. this basicaly was the reason for M2Ms instead of enum.

arturasmckwcz (approved):

PR #445, opened 2026-05-05
migrate legacy single-part district names to 12 official compound Berlin districts

nadavosa:
## Summary

- Adds a data migration that remaps all `location_district` rows referencing legacy sub-district names (e.g. `Kreuzberg`, `Moabit`, `Prenzlauer Berg`) to the correct compound Berlin district (e.g. `Friedrichshain-Kreuzberg`, `Mitte`, `Pankow`)
- Removes duplicate rows where a location already has the compound district linked
- Deletes the leftover legacy `district` records

Fixes the district filter returning 0 results and volunteer profiles showing old district names.

## Closes

https://github.com/need4deed-org/be/issues/442

## Test plan

- [ ] Run `yarn migration:run` on a local DB seeded with legacy data
- [ ] Verify volunteers previously showing `Kreuzberg`, `Moabit`, `Prenzlauer Berg` etc. now show the correct compound district
- [ ] Verify filtering by `Charlottenburg-Wilmersdorf`, `Friedrichshain-Kreuzberg` etc. returns the expected volunteers
- [ ] Verify no `location_district` rows reference non-existent districts after migration

### DIFF
--- a/src/data/migrations/1777973964566-consolidate-legacy-districts.ts
+++ b/src/data/migrations/1777973964566-consolidate-legacy-districts.ts
@@ -1,0 +1,67 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+const LEGACY_TO_COMPOUND: Record<string, string> = {
+  Kreuzberg: "Friedrichshain-Kreuzberg",
+  Friedrichshain: "Friedrichshain-Kreuzberg",
+  Charlottenburg: "Charlottenburg-Wilmersdorf",
+  Wilmersdorf: "Charlottenburg-Wilmersdorf",
+  Zehlendorf: "Steglitz-Zehlendorf",
+  Steglitz: "Steglitz-Zehlendorf",
+  Tempelhof: "Tempelhof-Schöneberg",
+  Schöneberg: "Tempelhof-Schöneberg",
+  Treptow: "Treptow-Köpenick",
+  Köpenick: "Treptow-Köpenick",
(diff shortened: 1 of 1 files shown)


src/data/migrations/1777973964566-consolidate-legacy-districts.ts
```diff
@@ -0,0 +1,67 @@
...
+            WHERE ld2.location_id = ld.location_id
+              AND ld2.district_id = (SELECT id FROM district WHERE title = $1)
+          )
+        `,
+        [compound, legacy],
+      );
+
+      // Remove any remaining rows still pointing to the legacy district
+      // (duplicates that were skipped above because the compound already existed).
+      await queryRunner.query(
```
arturasmckwcz, line 48: won't it produce `location` records that has no district linked by `location_district`?

nadavosa: Good catch — but the flow is safe. Step 1 (UPDATE) remaps the legacy row to the compound district. After that runs, the row no longer has the legacy district_id, so Step 2's DELETE (which targets rows still pointing to the legacy district_id) only hits rows that were NOT remapped — i.e. duplicates where the compound district was already present for that location. Those are safe to remove.

The only scenario where a location could end up with no district is if the compound district doesn't exist in the `district` table at all, which can't happen here since the 12 compound districts are always seeded before volunteers or opportunities are created.

PR #460, opened 2026-05-05
fix: include phone number in opportunity contact comment

nadavosa:
## Summary

- Adds `rac_phone` to the `<|>`-delimited contact comment saved when a new opportunity is submitted via the public form
- The phone number was already present in the form payload (`OpportunityLegacyFormData.rac_phone`) but was silently dropped when the comment was constructed

## Root cause

`legacy.routes.ts:85` built the comment text from `rac_email`, `rac_full_name`, `rac_address`, and `rac_plz` but omitted `rac_phone`.

## Closes

https://github.com/need4deed-org/be/issues/444

## Note for FE

Once this is deployed, the FE comment parser should be verified to render the `rac_phone` field correctly (the `<|>` delimiter count increases by one). Check https://github.com/need4deed-org/fe/issues/444 if tracking exists.

### DIFF
--- a/src/server/routes/opportunity/legacy.routes.ts
+++ b/src/server/routes/opportunity/legacy.routes.ts
@@ -82,7 +82,7 @@ export default async function opportunityLegacyRoutes(
       const commentRepository = fastify.db.commentRepository;
       await commentRepository.save(
         new Comment({
-          text: `${request.body.rac_email}<|>${request.body.rac_full_name}<|>${request.body.rac_address}<|>${request.body.rac_plz}`,
+          text: `${request.body.rac_email}<|>${request.body.rac_full_name}<|>${request.body.rac_address}<|>${request.body.rac_plz}<|>${request.body.rac_phone}`,
           entityId: id,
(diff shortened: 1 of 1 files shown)


nadavosa:
## Review

**Overview:** Appends `rac_phone` as the fifth field in the `<|>`-delimited contact comment saved when a legacy opportunity is submitted.

### What's good
- The FE renders this comment as plain text (no positional parsing) — confirmed by checking the FE codebase. Adding a field at the end is backward-compatible with existing stored comments.

### Issue — `rac_phone` can be `null`
The FE sends `rac_phone: string | null`. If the field is absent or empty, the comment will be saved as `"...<|>null"` (literal string). Guard it:
```ts
text: `${request.body.rac_email}<|>${request.body.rac_full_name}<|>${request.body.rac_address}<|>${request.body.rac_plz}<|>${request.body.rac_phone ?? ""}`,
```

One-character fix, but it matters for clean display — existing comments render the raw string as-is.

Otherwise LGTM.

PR #463, opened 2026-05-05
fix: map districtId in parseAgentPatch so agent district persists

nadavosa:
## Summary

- `parseAgentPatch` did not map `districtId`, so the agent's district was silently dropped on every PATCH
- The `Agent` entity has a `districtId` FK column that TypeORM persists directly — just needed to be included in the returned partial
- Uses a local type extension (`ApiAgentPatchExtended`) to expose `districtId` until it is added to `ApiAgentPatch` in the SDK (tracked in https://github.com/need4deed-org/sdk/issues/93); the local extension can be removed once the SDK publishes the updated type

## Out of scope

Agent `address` saving requires coordinating with the FE on the field format (structured vs. string) — deferred to a follow-up issue once the FE form is wired to the agent PATCH endpoint.

## Closes

https://github.com/need4deed-org/be/issues/451

## Depends on

https://github.com/need4deed-org/sdk/issues/93 (SDK PR should land first so the local type extension can be removed)


need4deed (approved):
.

PR #468, opened 2026-05-06
467 add logic for district setting to opp

### LINKED ISSUE #467: add logic for district setting to opp
### Description

add  a fn which calculates district for opportunity

### Proposed solution

_No response_

arturasmckwcz:
## Description

## Related Issues
Closes #467 

## Changes

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/data/utils/get-district.ts
+++ b/src/data/utils/get-district.ts
@@ -1,3 +1,4 @@
+import { Voidable } from "../../server/utils";
 import { dataSource } from "../data-source";
 import District from "../entity/location/district.entity";
 import Postcode from "../entity/location/postcode.entity";
@@ -5,7 +6,7 @@ import DistrictPostcode from "../entity/m2m/district-postcode";
 import { getRepository } from "./get-repository";
 
 export async function getDistrictFromPostcode(
-  postcode: Postcode,
+  postcode: Voidable<Postcode>,
 ): Promise<District | null> {
   if (postcode?.id) {
     const districtPostcodeRepository = getRepository(
--- a/src/server/routes/opportunity/legacy.routes.ts
+++ b/src/server/routes/opportunity/legacy.routes.ts
@@ -11,10 +11,8 @@ import {
 import { ILike, In } from "typeorm";
 import { UnauthorizedError } from "../../../config";
 import Comment from "../../../data/entity/comment.entity";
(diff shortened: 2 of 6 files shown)


nadavosa (commented):
The approach follows the existing agent-district pattern well, but there's one bug and a couple of smaller things worth addressing.

**Bug — missing `await` in the GET list handler (`opportunity.routes.ts`)**

```ts
// line ~207 – inside opportunities.map()
addDistrictToOpportunity(opportunity),   // ← no await
```

`addDistrictToOpportunity` is `async`, so without `await` this returns a `Promise` object. `Object.assign` will merge the Promise's own properties (none that are useful) into `opportunity.deal.profile` instead of the resolved opportunity, so the district is never set and `opportunityUpdates` will always be empty. The save call at the end becomes a no-op.

The GET single-opportunity handler above it correctly uses `await` — the list handler needs the same treatment. Since `Array.map` doesn't support async callbacks, it needs to become a `Promise.all`:

```ts
const opportunitiesCategory = await Promise.all(
  opportunities.map(async (opportunity) => {
    Object.assign(
      opportunity.deal.profile,
      addCategoryToProfile(opportunity.deal.profile),
    );
    await addDistrictToOpportunity(opportunity);
    return opportunity;
  }),
);
```

Note also that `addDistrictToOpportunity` mutates `opportunity` in-place before returning it, so the `Object.assign` target should be `opportunity`, not `opportunity.deal.profile` (same as in the single-get handler).

---

**Non-null assertion on optional chain (`add-district-to-opp.ts`)**

```ts
opportunity.agent?.address?.postcode!
```

The `?` chains already propagate `undefined` if any link is missing. The trailing `!` asserts non-null but the value can still be `undefined` — it just suppresses the compiler warning. `getDistrictFromPostcode` guards with `if (postcode?.id)` so it handles `undefined` fine. Drop the `!`.

---

**Logic duplication in `legacy.routes.ts`**

The legacy route re-implements the district-from-agent and district-from-postcode logic inline rather than calling `getDistrictToOpportunityHandler`. If the lookup logic changes later it'll need updating in two places. Consider using the shared handler there too.

nadavosa (approved):
LGTM — all review issues addressed.

PR #473, opened 2026-05-07
470 add district to schemas and dtos

### LINKED ISSUE #470: add district to schemas and dtos
### Description

update SDK types, DTOs, and schemas with district

### Proposed solution

_No response_

arturasmckwcz:
## Description

## Related Issues
Closes #470 

## Changes

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "fastify": "^5.3.3",
     "fastify-mailer": "^2.3.1",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "^0.0.81",
+    "need4deed-sdk": "^0.0.82",
     "nodemailer": "^7.0.4",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
--- a/src/data/utils/get-district.ts
+++ b/src/data/utils/get-district.ts
@@ -23,3 +23,13 @@ export async function getDistrictFromPostcode(
   }
   return null;
 }
+
+export async function getDistrictByTitle(
+  title: string,
+): Promise<District | null> {
+  const districtRepository = getRepository(dataSource, District);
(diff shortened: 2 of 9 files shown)


nadavosa (commented):
Reviewed — a few notes before approve.

**`add-district-to-opp.ts`** — the non-accompanying branch reads `opportunity.deal?.location?.locationDistrict?.[0]?.district`. This works only if the opportunity is fetched with the full `deal.location.locationDistrict.district` relation chain. Can you confirm those relations are included in the existing fetch that calls this handler? If not, this will silently no-op in prod.

**`parser-opportunity-legacy.ts`** — `getDistrictByTitle(body.berlin_locations?.[0] ?? "")` passes an empty string when the field is absent. `findOne({ where: { title: "" } })` returns null safely, so it's not a bug, but `body.berlin_locations?.[0] ? getDistrictByTitle(...) : null` would be more explicit.

**`dto-opportunity.ts`** — the `as ApiOpportunityGetList` / `as ApiOpportunityGet` casts suppress a TS error rather than fixing it (the spread pattern loses type info). It's acceptable here since the shape is correct, just worth knowing this is a cast-to-satisfy, not a type-safe construction.

**Heads up on #474** — that PR refactors `deal.profile.*` → `deal.*` throughout `dto-opportunity.ts`. The two PRs will conflict on merge. Merge order: #473 first, then #474 rebases on top.

Everything else looks good — SDK bump correct, `parseOpportunityLegacy` correctly awaited in the route, tests cover the new non-accompanying path. Just confirm the relation chain and I'll approve.

need4deed (approved):

PR #486, opened 2026-05-11
fix: remove condition for assigning from locations

arturasmckwcz:
## Description

## Related Issues

## Changes
- change district assignment logic: remove condition if only not accompanying

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/server/utils/data/add-district-to-opp.ts
+++ b/src/server/utils/data/add-district-to-opp.ts
@@ -1,4 +1,4 @@
-import District from "../../../data/entity/location/district.entity";
+import { OpportunityType } from "need4deed-sdk";
 import Opportunity from "../../../data/entity/opportunity/opportunity.entity";
 import { getDistrictFromPostcode } from "../../../data/utils/get-district";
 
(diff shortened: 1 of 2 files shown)


nadavosa (commented):
## Logic issue: accompanying opportunities get wrong district source

The PR removes the `type !== ACCOMPANYING` guard so all opportunity types now attempt to get their district from `deal.location.locationDistrict`. This is wrong for accompanying opportunities.

**For accompanying opps**, the location is the appointment address filled in the form — stored as `deal.postcode` (the `Postcode` entity on `Deal`). The correct source is:

```ts
const district = await getDistrictFromPostcode(opportunity.deal?.postcode);
```

**For regular opps**, `deal.location.locationDistrict` is correct — no change needed there.

The original type guard was right to separate the two paths; the problem was only which postcode source was used for accompanying.

Correct logic:
```ts
if (opportunity.type !== OpportunityType.ACCOMPANYING) {
  const district = opportunity.deal?.location?.locationDistrict?.[0]?.district;
  if (district) {
    opportunity.district = district;
    updates.push(opportunity);
  }
  return opportunity;
}
// accompanying: use appointment postcode from the form
const district = await getDistrictFromPostcode(opportunity.deal?.postcode);
if (district) {
  opportunity.district = district;
  updates.push(opportunity);
  return opportunity;
}
// fallback to agent
if (opportunity.agent?.districtId) {
  opportunity.districtId = opportunity.agent.districtId;
  updates.push(opportunity);
  return opportunity;
}
const districtFromAgent = await getDistrictFromPostcode(opportunity.agent?.address?.postcode);
if (districtFromAgent) {
  opportunity.district = districtFromAgent;
  updates.push(opportunity);
}
return opportunity;
```

nadavosa (changes requested):
Good direction on the accompanying postcode logic — that part is correct. Two issues to fix before merging:

## 1. Critical: early-return guard was removed

```ts
if (!opportunity || opportunity.districtId) {
  return opportunity;
}
```

This guard was removed but it does two important things:
- **null safety**: without it, `opportunity.type` on the next line throws a TypeError if called with `undefined`
- **idempotency**: without it, opportunities that already have a `districtId` get re-processed and their district can be silently overwritten

Please restore this as the very first check inside `addDistrictToOpportunity`.

## 2. Tests for those cases were deleted

The two deleted test cases (`"returns undefined when opportunity is falsy"` and `"returns the opportunity unchanged when it already has a districtId"`) covered real behaviour that still needs to exist. Restore them once the guard is back.

PR #490, opened 2026-05-11
add: early exit if no logic needed

arturasmckwcz:
## Description

## Related Issues

## Changes
- add early exit if no logic needed

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/data/utils/get-district.ts
+++ b/src/data/utils/get-district.ts
@@ -8,13 +8,21 @@ import { getRepository } from "./get-repository";
 export async function getDistrictFromPostcode(
   postcode: Voidable<Postcode>,
 ): Promise<District | null> {
-  if (postcode?.id) {
+  let postcodeId = postcode?.id;
+  if (!postcodeId && postcode?.value) {
+    const postcodeRepository = getRepository(dataSource, Postcode);
+    const postcodeEntity = await postcodeRepository.findOne({
+      where: { value: postcode.value },
+    });
+    postcodeId = postcodeEntity?.id;
+  }
+  if (postcodeId) {
     const districtPostcodeRepository = getRepository(
       dataSource,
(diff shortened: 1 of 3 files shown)


nadavosa (approved):
The guard is correct and useful — avoids unnecessary async DB lookups () on opportunities that already have a district set, and prevents accidentally overwriting a correctly-set district on re-runs.

One thing to keep in mind: if a district was set incorrectly during creation (e.g. from bad postcode data), this guard would permanently block auto-correction on any future re-run of this function. That seems like an acceptable trade-off — incorrect districts should be fixed explicitly rather than silently overwritten — but worth being aware of.

No tests added, but the logic is simple enough that's fine here.

LGTM, safe to merge.

nadavosa (approved):
Reviewed the full diff now (my earlier review only covered add-district-to-opp.ts).

The `get-district.ts` change is the real core fix here: it adds a DB lookup by postcode value when the Postcode object has no id. This was the root cause — `new Postcode({ value: body.accomp_postcode })` produces an object with no id, so the old `if (postcode?.id)` guard short-circuited and returned null every time.

The three changes compose correctly:
1. `get-district.ts` — resolves district from a value-only postcode by looking up the id first
2. `parser-opportunity-legacy.ts` — sets `districtId: district?.id` directly (cleaner than setting the full relation) so new opportunities created via the form get the district set immediately
3. `add-district-to-opp.ts` — early exit prevents overwriting an already-set districtId on subsequent runs

Old opportunities (created before this fix, districtId = null) are unaffected by the early exit because their Deal postcode is loaded from the DB with a real id — the original code path still runs for them.

LGTM on the full diff, this is the right fix.

PR #510, opened 2026-05-12
restore: postcode relation on Accompanying entity (re-do of #502/#503)

arturasmckwcz:
Re-applies PR #502/#503's postcode work after develop was rewritten. PII-laden seed migrations from the original branch are **not** included — only the structural/postcode pieces.

## Summary

- Add `postcode` ManyToOne relation to `Accompanying` entity (FK `SET NULL`)
- Add migration `1778578280034-add-postcode-to-accompanying`
- Add migration `1778579676871-update-accompanying-postcode-fk-set-null`
- Move `getPostcode` helper to `src/data/utils/` (fixes services → server import direction); re-export via `for-routes.ts` for existing call sites
- Update `parser-accompanying-legacy.ts` to resolve `accomp_postcode` to a `Postcode` entity
- Add unit tests for `accompanyingParserOpportunity`

## Scope note

`parser-accompanying-legacy.ts` also restores the `parseAccompDatetime` helper from PR #501 (Europe/Berlin parsing). The postcode commit was originally based on top of #501; preserving the function shape required keeping that helper. Small incidental restoration.

## Known pre-existing issue on develop (not introduced here)

`yarn typecheck` reports two errors in `dto-accompanying.ts` and `parser-opportunity-patch-data.ts` about `languageToTranslate` not existing on `ApiOpportunityAccompanyingDetails`. These exist on `develop` today because the installed SDK is 0.0.83 but the code still uses pre-0.0.83 field names. The follow-up PR (re-do of #506) fixes this.

## Test plan

- [ ] Migrations run cleanly (`yarn migration:run`)
- [ ] POST legacy accompanying form with `accomp_postcode` resolves to the right `Postcode` row
- [ ] Deleting a postcode used by an accompanying sets the FK to NULL (does not cascade-delete the accompanying)

### DIFF
--- a/src/data/entity/location/postcode.entity.ts
+++ b/src/data/entity/location/postcode.entity.ts
@@ -1,11 +1,11 @@
 import { IsNotEmpty } from "class-validator";
 import { Column, Entity, OneToMany, PrimaryGeneratedColumn } from "typeorm";
-
 import { IsPostcode } from "../../../services/validators/custom";
 import { Country, GermanCity } from "../../types";
 import Deal from "../deal.entity";
 import DistrictPostcode from "../m2m/district-postcode";
 import LocationPostcode from "../m2m/location-postcode";
+import Accompanying from "../opportunity/accompanying.entity";
 import Address from "./address.entity";
 
 @Entity()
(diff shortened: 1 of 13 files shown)


need4deed (approved):
LGTM — restore postcode infrastructure (no PII seed data).

PR #518, opened 2026-05-13
feat: emit appointmentPostcode and appointmentDistrict in GET /opportunity/:id

arturasmckwcz:
Closes #517

## Summary

- Bump \`need4deed-sdk\` to \`^0.0.84\` (adds \`appointmentPostcode\` and \`appointmentDistrict\` to \`ApiOpportunityAccompanyingDetails\`)
- Emit \`appointmentPostcode\` from \`accompanying.postcode\` (already loaded as a relation by the route)
- Derive \`appointmentDistrict\` via the existing \`getDistrictFromPostcode\` helper in the route handler, then pass it into \`dtoOpportunityGet\` / \`dtoOpportunityAccompanying\` (keeps the DTO a pure sync transform)
- Add \`appointmentPostcode\` and \`appointmentDistrict\` to the \`ApiAccompanying\` response schema in \`sdk-types.json\` so Fastify doesn't strip them

## Scope note

Only \`GET /opportunity/:id\` is updated. The list endpoints (\`dtoOpportunityGetList\`, \`dtoVolunteerOpportunityGetList\`) still call \`dtoOpportunityAccompanying\` without a district argument — they continue to return the same shape they did before. If the FE needs these two fields in list responses too, that can be a follow-up.

## Test plan

- [ ] \`GET /opportunity/:id\` for an accompanying opportunity with a postcode returns \`appointmentPostcode: { id }\`
- [ ] When that postcode maps to a district, the response also includes \`appointmentDistrict: { id }\`
- [ ] When there's no postcode, neither field is present
- [ ] \`yarn typecheck\` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "fastify": "^5.3.3",
     "fastify-mailer": "^2.3.1",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "^0.0.83",
+    "need4deed-sdk": "^0.0.84",
     "nodemailer": "^7.0.4",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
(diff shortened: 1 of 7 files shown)


need4deed (approved):
LGTM — schema, DTO and test coverage all in sync with SDK 0.0.84.

PR #520, opened 2026-05-13
fix: persist appointmentPostcode and refugeeLanguage on PATCH; reject appointmentDistrict

arturasmckwcz:
Closes #519.

## Problem

\`PATCH /volunteer/opportunity/:id\` validated request bodies against \`ApiAccompanying\` (the response shape), so clients could send fields the parser silently dropped:

- \`refugeeLanguage\` — accepted since #506, never persisted
- \`appointmentPostcode\` — accepted since #518, never persisted
- \`appointmentDistrict\` — accepted since #518; cannot meaningfully be persisted (district is derived from postcode on the response side)

### Schema split (\`sdk-types.json\`)

- New \`ApiAccompanyingPatch\` — subset of \`ApiAccompanying\` without \`appointmentDistrict\`, with \`additionalProperties: false\` for strict request validation.
- \`ApiVolunteerOpportunityPatch.accompanyingDetails\` retargeted to the new schema.
- GET responses continue to use \`ApiAccompanying\` (unchanged).

### Parser (\`parser-opportunity-patch-data.ts\`)

- \`appointmentPostcode.id\` is now set as a \`postcode: { id }\` relation on the \`Accompanying\` patch. TypeORM \`.update()\` translates this to the \`postcode_id\` FK column; an invalid id surfaces as 400 via \`patchEntity\`'s \`QueryFailedError\` handler.
- \`refugeeLanguage\` is folded into the \`languages\` array with \`LangPurpose.TRANSLATION\`, deduplicated against \`languagesResidents\` (same (id, purpose) combinations collapsed to one).
- Same id can still coexist under different purposes — e.g. someone is both a main language speaker and a translation-target — which is the existing model.

## SDK type vs. backend schema

The SDK type \`ApiOpportunityPatch.accompanyingDetails: ApiOpportunityAccompanyingDetails\` still type-checks as if \`appointmentDistrict\` were writable. The backend now rejects it at runtime. SDK side can publish a follow-up \`ApiOpportunityAccompanyingDetailsPatch\` type if needed; not blocking.

## Tests

\`src/test/services/dto/parser-opportunity-patch-data.test.ts\` — 10 cases:

- Throws on falsy body
- Empty body → \`null\` for \`accompanying\` and \`languages\` (via \`getEmptyPropsNull\`)
- Base accompanying fields map correctly
- \`appointmentDate\` + \`appointmentTime\` compose into \`Date\`
- \`appointmentPostcode\` → \`postcode: { id }\` relation set
- Missing \`appointmentPostcode\` → no \`postcode\` in patch
- \`refugeeLanguage\` → \`TRANSLATION\` purpose
- Dedupe vs. \`languagesResidents\` (same id, same purpose)
- Same id under different purposes coexists (\`GENERAL\` + \`TRANSLATION\`)

## Test plan

- [ ] PATCH with \`appointmentDistrict\` in body → 400
- [ ] PATCH with \`appointmentPostcode: { id }\` → postcode_id updated on the accompanying row
- [ ] PATCH with \`refugeeLanguage: [{id}]\` and no \`languagesResidents\` → ProfileLanguage rows created/diffed with TRANSLATION purpose
- [ ] PATCH with both \`refugeeLanguage\` and \`languagesResidents\` overlapping → no duplicate ProfileLanguage rows

### DIFF
--- a/src/server/plugins/typeorm.ts
+++ b/src/server/plugins/typeorm.ts
@@ -6,6 +6,7 @@ import Comment from "../../data/entity/comment.entity";
 import Communication from "../../data/entity/communication.entity";
 import Document from "../../data/entity/document.entity";
 import FieldTranslation from "../../data/entity/field_translation.entity";
+import Postcode from "../../data/entity/location/postcode.entity";
 import OpportunityVolunteer from "../../data/entity/m2m/opportunity-volunteer";
 import Agent from "../../data/entity/opportunity/agent.entity";
 import Opportunity from "../../data/entity/opportunity/opportunity.entity";
@@ -47,6 +48,7 @@ const typeormPlugin: FastifyPluginAsync = async (fastify) => {
       profileRepository: dataSource.getRepository(Profile),
       agentRepository: dataSource.getRepository(Agent),
       organizationRepository: dataSource.getRepository(Organization),
+      postcodeRepository: dataSource.getRepository(Postcode),
     });
 
     // TODO: add validation of others
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -306,6 +306,19 @@ export default async function opportunityRoutes(
       }
 
       if (accompanying) {
+        const appointmentPostcodeId =
+          request.body.accompanyingDetails?.appointmentPostcode?.id;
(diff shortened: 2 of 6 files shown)


need4deed (approved):
LGTM — schema split + Postcode resolution in the route handler are the right shape. Follow-up #519-style work on updateOptionList destructive semantics can come separately.

PR #528, opened 2026-05-18
test(#223): add unit tests for data utils

### LINKED ISSUE #223: add data utils tests
### Description

add unit tests for `/src/data/utils`

### Proposed solution

_No response_

nadavosa:
## Summary
- `get-rrule.test.ts` — all 7 weekdays (case-insensitive), invalid names, non-string input, whitespace edge case
- `get-start-end-dates.test.ts` — `getStartEndDates` hour/minute output; `getStartEnd` label-to-hour mapping using `.getHours()` assertions (timezone-safe)
- `passwd.test.ts` — mocks `bcrypt` to test call arguments and error-wrapping behaviour without slow 10-round hashing
- `fetch-json.test.ts` — file-path vs URL routing, explicit `fs/promises` factory mock for reliable Node built-in interception

## Test plan
- [ ] Run `yarn test:run src/test/data/utils/` and confirm all tests pass without a DB connection

Closes https://github.com/need4deed-org/be/issues/223


arturasmckwcz (commented):
Thanks @nadavosa — `getRRULE` and `passwd` suites are accurate and well-targeted (non-string guard, no-trim whitespace, case-insensitivity; salt-rounds + both error wrappers). Reviewed each assertion against the source. A few notes, mostly nits:

**1. (medium) `getStartEndDates` validation is dead code — the tests don't reveal it.** The guard `startHour < endHour && startHour < 0 && startHour > 24 && endHour < 0 && endHour > 24` is unsatisfiable, so invalid hours never throw. The suite only covers valid inputs, so it implies validation works when it doesn't. Either fix the guard (`startHour > endHour || startHour < 0 || endHour > 24`) and test the throw, or add a test asserting invalid input currently does **not** throw, so the gap is explicit. (Happy to land a tiny follow-up for the guard.)

**2. (low) `get-start-end-dates.test.ts` codifies a wrong map entry.** `getStartEnd`'s map has incorrect values (e.g. `"8:00 - 10:00" → 8..11`, and `"17:00-20:00" → 14..17`). The test asserts the `8:00-10:00 → 8..11` value as expected — fine as characterization (and the comment is honest), but it locks the bug in. Worth fixing the source map separately and updating the expectation.

**3. (low) timezone-fragile date assertion.** `expect(end.getFullYear()).toBe(2024)` where `end = new Date(\"2024-01-01\")` (ISO date-only → parsed as UTC). In a UTC-negative TZ that's 2023 and the test fails. Green in Berlin/UTC (project default), but brittle. The `getHours()` assertions are fine (local-time date form).

**4. (low) leaked global stub.** `fetch-json.test.ts` calls `vi.unstubAllGlobals()` inline; if an assertion throws first, the `fetch` stub leaks into later tests. Move to `afterEach(() => vi.unstubAllGlobals())`.

**5. (low)** branch is behind `develop` (forked ~#578) — test-only, so low risk, but a rebase before merge runs it against current utils.

Approvable after the nits; #1 is the one worth addressing (it's a real source bug the tests currently paper over).

need4deed (approved):

PR #534, opened 2026-05-18
fix(#282): return 400 for invalid status/type enum params on volunteer opportunity list

### LINKED ISSUE #282: wrong http status code for invalid data
### Summary

now API responds with 500 if wrong column value is sent

### Steps to Reproduce

1. send `<API base url>/volunteer/6/opportunity?page=20&limit=6&status=post`

### Expected Behavior

responds with 400 "bad request"

### Actual Behavior

responds with 500 "Something went wrong."

### Environment

_No response_

nadavosa:
## Summary
- Validates `status` and `type` query params against their respective enums before passing to TypeORM
- Throws `BadRequestError` (400) instead of letting an invalid value reach the DB and cause a 500

## Related
- Closes https://github.com/need4deed-org/be/issues/282

## Test plan
- [ ] `GET /volunteer/6/opportunity?status=post` returns 400 with a descriptive message
- [ ] `GET /volunteer/6/opportunity?status=opp-active&type=vol-regular` returns 200


src/server/routes/volunteer/volunteer-opportunity.routes.ts
arturasmckwcz, line 30: why do we need code below for validation. pls update querystring schema instead

nadavosa (commented):
Addressed — removed the manual `BadRequestError` guards for `status` and `type`. The AJV schema (`volunteerOpportunityListQuerySchema`) already validates both fields against their respective enums via `getRef()`, so the route-level checks were redundant. Also rebased onto develop.

PR #540, opened 2026-05-19
feat(#539): accept agent.id on PATCH /opportunity/{id} to re-link agent

arturasmckwcz:
Closes #539.

## Summary
- `PATCH /opportunity/{id}` now accepts `{ agent: { id: <existing_id> } }` as an explicit re-link reference; the opportunity's `agentId` is updated to point at the existing agent without mutating the agent's name/address/district.
- The legacy `{ name, address, district }` shape still works for in-place edits (recommendation per the issue: deprecate once the frontend cuts over).
- When `agent.id` is present, it wins — other fields on `agent` are ignored.
- Swagger (`/swagger/json`) now reflects `id` on `ApiVolunteerOpportunityPatch.agent`.

## Implementation
- `src/server/schema/sdk-types.json` — added `id: integer` to `ApiVolunteerOpportunityPatch.agent`.
- `src/services/dto/parser-opportunity-patch-data.ts` — new local types `OpportunityAgentPatchBody` / `OpportunityPatchBody` (the SDK type is stale on `agent` shape); parser surfaces `agentLinkId` and skips the in-place agent patch when an id is sent.
- `src/server/routes/opportunity/opportunity.routes.ts` — when `agentLinkId` is set, verify the agent exists (400 if not) and update `opportunity.agentId`; otherwise fall back to the existing in-place agent patch.
- `src/test/services/dto/parser-opportunity-patch-data.test.ts` — 3 new parser tests (relink, legacy in-place edit, id-wins).

## Test plan
- [x] `yarn typecheck`
- [x] `yarn test:run` (161/161 pass)
- [ ] Manual: `PATCH /opportunity/<id>` with `{ "agent": { "id": <other_agent_id> } }` → 200; subsequent `GET` shows the new agent inlined; the previously-linked agent record is untouched.
- [ ] Manual: `PATCH /opportunity/<id>` with `{ "agent": { "id": 99999999 } }` → 400 `Agent (id:99999999) not found.`
- [ ] Manual: `PATCH /opportunity/<id>` with the legacy `{ "agent": { "name": "..." } }` shape still patches the title in place.
- [ ] Confirm Swagger UI exposes the new `id` field on the agent body.

### DIFF
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -2,6 +2,7 @@ import { FastifyInstance, FastifyPluginOptions } from "fastify";
 import {
   ApiOpportunityGet,
   ApiOpportunityGetList,
+  ApiOpportunityPatch,
   SortOrder,
   UserRole,
 } from "need4deed-sdk";
@@ -21,7 +22,6 @@ import logger from "../../../logger";
 import {
   dtoOpportunityGet,
   dtoOpportunityGetList,
-  OpportunityPatchBody,
   parseOpportunity,
 } from "../../../services";
 import {
@@ -238,7 +238,7 @@ export default async function opportunityRoutes(
 
   fastify.patch<{
     Params: ParamsId;
-    Body: OpportunityPatchBody;
+    Body: ApiOpportunityPatch;
     Reply: ReplyMessage;
   }>(
     "/:id",
@@ -269,13 +269,13 @@ export default async function opportunityRoutes(
         opportunity: opportunityObj,
         contact,
         agent,
-        agentLinkId,
         accompanying,
         languages,
         schedule,
         skills,
         activities,
       } = parseOpportunity(request.body);
+      const agentLinkId = request.body.agent?.id;
       logger.debug(
         `PATCH /opportunity/{id} ${JSON.stringify(parseOpportunity(request.body))}`,
       );
@@ -305,7 +305,7 @@ export default async function opportunityRoutes(
           where: { id: agentLinkId },
         });
         if (!linkedAgent) {
-          throw new BadRequestError(`Agent (id:${agentLinkId}) not found.`);
+          throw new NotFoundError(`Agent (id:${agentLinkId}) not found.`);
         }
         const success = await patchEntity(
           Opportunity,
@@ -313,7 +313,7 @@ export default async function opportunityRoutes(
           opportunity.id,
         );
         if (!success) {
-          throw new Error("Relinking opportunity agent failed.");
+          throw new BadRequestError("Relinking opportunity agent failed.");
         }
       } else if (agent) {
         const success = await patchEntity(Agent, agent, opportunity.agentId);
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -2,7 +2,6 @@ import { FastifyInstance, FastifyPluginOptions } from "fastify";
 import {
   ApiOpportunityGet,
   ApiOpportunityGetList,
-  ApiOpportunityPatch,
   SortOrder,
   UserRole,
 } from "need4deed-sdk";
@@ -22,6 +21,7 @@ import logger from "../../../logger";
 import {
   dtoOpportunityGet,
   dtoOpportunityGetList,
+  OpportunityPatchBody,
   parseOpportunity,
 } from "../../../services";
 import {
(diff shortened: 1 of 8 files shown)


nadavosa (commented):
**Overall: looks good to merge.** Logic is correct, backward-compatible, and well-tested.

---

**One real issue — `address` and `district` silently ignored in in-place edit:**

The parser only maps `agentBody.name → title`. The `address` and `district` fields exist in `OpportunityAgentPatchBody` but are never applied to the Agent patch:

\`\`\`ts
agent: agentBody && agentLinkId === undefined
  ? ({ title: agentBody.name } as Partial<Agent>)  // address, district dropped
  : {},
\`\`\`

This was the same before this PR, so it is pre-existing — but since we are touching this code it is worth calling out. A coordinator editing the RAC address via the form today would see their change silently lost. Worth a follow-up issue if not fixing here.

---

**Minor observations:**

- `{ agentId: agentLinkId } as Partial<Opportunity>` — the cast is needed because TypeORM's partial entity type does not expose FK columns directly. Fine as-is, just worth a comment.
- The `OpportunityStatusType` enum reformatting in `sdk-types.json` is cosmetic and harmless.
- The local `OpportunityPatchBody` workaround for the stale SDK type is acceptable short-term — should go into a future SDK update.
- All 3 parser test cases are exactly the right ones to have (relink, legacy, id-wins).

---

**Verdict:** Approve with the note that the silent address/district drop should be tracked in a follow-up issue. The relink path is correct and production-safe.

src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -299,7 +300,22 @@ export default async function opportunityRoutes(
...
+        if (!linkedAgent) {
+          throw new BadRequestError(`Agent (id:${agentLinkId}) not found.`);
+        }
+        const success = await patchEntity(
+          Opportunity,
+          { agentId: agentLinkId } as Partial<Opportunity>,
+          opportunity.id,
+        );
+        if (!success) {
+          throw new Error("Relinking opportunity agent failed.");
```
arturasmckwcz: throwing  new Error returns 500 which is semantically wrong, isn't it? 400 would be more appropriate. wouldn't use custom `BadRequestError` be more to the point? or maybe `NotFoundError`?

src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -238,7 +238,7 @@ export default async function opportunityRoutes(
 
   fastify.patch<{
     Params: ParamsId;
-    Body: ApiOpportunityPatch;
+    Body: OpportunityPatchBody;
```
arturasmckwcz: I understand that JSON schema used by fastify is the main contract, but SDK lifts shared structures to transpilation phase which is worth bothering IMHO

let's do `ApiOpportunityPatch` refactoring instead!

src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -269,6 +269,7 @@ export default async function opportunityRoutes(
         opportunity: opportunityObj,
         contact,
         agent,
+        agentLinkId,
```
arturasmckwcz: why move agent id outside agent object? wouldn't agent.id do the thing?

src/server/routes/opportunity/opportunity.routes.ts
arturasmckwcz: Fixed in e020286: missing referenced agent now throws `NotFoundError` (404), and a failed relink update throws `BadRequestError` (400). No more generic 500.

src/server/routes/opportunity/opportunity.routes.ts
arturasmckwcz: Agreed and done. SDK 0.0.86 adds `agent.id` (and deprecates `name`/`address`/`district`) on `ApiOpportunityPatch`. Bumped the dependency, removed the local `OpportunityPatchBody` workaround, and the route now types its body as `ApiOpportunityPatch` again.

src/server/routes/opportunity/opportunity.routes.ts
arturasmckwcz: Good call. Removed the hoisted `agentLinkId` from the parser; the route now reads `request.body.agent?.id` directly. The parser just suppresses the in-place agent patch when an id is present.

need4deed (approved):
Approved — review comments addressed, tests/typecheck/lint green. Follow-up tracked in #542.

PR #545, opened 2026-05-21
fix(#433): guard vol- prefix to prevent double-prefixing on engagement/match filters

nadavosa:
## Summary

Adds an idempotency check to the `vol-` prefix workaround in the volunteer list route.

Previously `.map((e) => `vol-${e}`)` would produce `vol-vol-xxx` if a value already carried the prefix (e.g. a direct API call or a client that pre-prefixes). Now the prefix is only added when not already present.

Applies to both `engagement` and `match` filter values.

> Note: the full match-filter feature (`statusMatch` in `getVolunteerWhere`, `match` in `QuerystringVolunteerFiltering`, schema update) was already merged into develop via earlier PRs.

Closes https://github.com/need4deed-org/be/issues/433

## Test plan

- [ ] `GET /volunteer?filter[engagement]=active` → `statusEngagement = vol-active` (prefix added)
- [ ] `GET /volunteer?filter[engagement]=vol-active` → `statusEngagement = vol-active` (no double prefix)
- [ ] Same for `filter[match]`
- [ ] Existing volunteer list without filters returns results as before

### DIFF
--- a/src/server/routes/volunteer/volunteer.routes.ts
+++ b/src/server/routes/volunteer/volunteer.routes.ts
@@ -157,11 +157,11 @@ export default async function volunteerRoutes(
           const engagement = [query.filter.engagement]
             .flat()
             .filter(Boolean)
-            .map((e) => `vol-${e}`);
+            .map((e) => (e.startsWith("vol-") ? e : `vol-${e}`));
           const match = [query.filter.match]
             .flat()
             .filter(Boolean)
-            .map((m) => `vol-${m}`);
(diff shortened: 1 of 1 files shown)


nadavosa (commented):
Rebased onto develop and resolved conflicts. The `matchWorkaround` logic was already merged into develop's `filterWorkaround` function via a later PR — kept that combined version and preserved the idempotency guard (`.startsWith("vol-")`) from this PR which was missing from develop.

PR #566, opened 2026-05-25
chore: sync entities with DDL — add video-call to communication contact method enum

arturasmckwcz:
## What

Syncs the database DDL with the entities. The `communication_contact_method_enum` entity had gained a `'video-call'` value that was never captured in a migration.

## Change

New migration `1779737272182-sync-entities-ddl` recreates `communication_contact_method_enum` to include `'video-call'` (Postgres enum rename-swap pattern, with a proper `down`).

Generated via `yarn migration:generate`. A follow-up `migration:generate` reports **"No changes in database schema were found"**, confirming entities and DDL are fully in sync.


need4deed (approved):
Approving: entity/DDL sync migration, no schema drift remaining.

PR #567, opened 2026-05-25
feat: add UNIQUE(opportunity_id, volunteer_id) to OpportunityVolunteer

arturasmckwcz:
## What

Adds a `@Unique(["opportunityId", "volunteerId"])` constraint to the `OpportunityVolunteer` entity so a volunteer can have at most one match record per opportunity.

## Pre-existing duplicates

The dev and **production** databases already contain rows that violate this. The migration removes them *before* adding the constraint, keeping exactly one row per `(opportunity_id, volunteer_id)` pair via a single window-function ranking:

- prefer `status = 'opp-matched'` when statuses differ
- otherwise keep the latest (`created_at DESC`, then `id DESC`)

In dev this deleted 4 violators across 4 groups; the correct survivors (2 matched, 2 latest-pending) remain. Verified afterwards: 0 duplicates, constraint present, no schema drift, `typecheck` passes.

### Edge case
A group with differing statuses but *no* `opp-matched` falls back to keeping the latest. None exist in dev; prod may differ.

### DIFF
--- a/src/data/entity/m2m/opportunity-volunteer.ts
+++ b/src/data/entity/m2m/opportunity-volunteer.ts
@@ -8,12 +8,14 @@ import {
   JoinColumn,
   ManyToOne,
   PrimaryGeneratedColumn,
+  Unique,
   UpdateDateColumn,
 } from "typeorm";
 import Opportunity from "../opportunity/opportunity.entity";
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Approving: unique constraint with safe pre-dedupe migration; verified no remaining duplicates.

PR #568, opened 2026-05-25
chore: seed matched OpportunityVolunteer records

arturasmckwcz:
## What

Seeds 189 `(volunteer, opportunity)` pairs as **matched**, implementing the provided seed logic as a migration.

A single idempotent upsert (enabled by the new `UNIQUE (opportunity_id, volunteer_id)` constraint):

- **insert** a matched row when none exists
- **update** `pending` → `matched` on conflict
- **leave** `matched` / `active` / `past` rows untouched (`ON CONFLICT ... WHERE` guard)

Pairs whose opportunity or volunteer no longer exists are **skipped** — the inner joins to `opportunity` and `volunteer` drop them, so a stale reference can't fail the deploy.

## Verified (dev DB)

Of 189 pairs, 184 were seedable → all now `opp-matched`, **0 still pending**; 5 pairs with a missing opportunity/volunteer were skipped. Migration is idempotent (safe to re-run); `down()` is a no-op (prior per-row statuses aren't recoverable).

### DIFF
--- a/src/data/migrations/1779739039589-seed-matched-opp-vol.ts
+++ b/src/data/migrations/1779739039589-seed-matched-opp-vol.ts
@@ -1,0 +1,240 @@
+import { OpportunityVolunteerStatusType } from "need4deed-sdk";
+import { MigrationInterface, QueryRunner } from "typeorm";
+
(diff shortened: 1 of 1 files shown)


need4deed (approved):
Approving: idempotent matched-seed migration; verified in dev, skips stale refs.

PR #574, opened 2026-05-26
chore(#573): remove unused volunteer_list_mv scaffolding

arturasmckwcz:
Closes #573. Closes #571.

## Why

The `volunteer_list_mv` materialized view was already dropped at the DB level by migration `1763036250587-add-config-drop-volunteer-list-mv.ts`, but the supporting code remained:

- an entity wrapper pretending to read from a non-existent view
- a `createVolunteerListMV` function that was never called from runtime code
- a TypeORM repository (`volunteerListMvRepository`) wired into the Fastify instance — pointing at a phantom table
- a filter utility (`getFilteredVolunteers`) that was never wired into a route and had field-name mismatches with the actual schema/type (the wire-up was tracked in #571)

The MV approach isn't part of the MVP. Performance optimizations like this can return post-MVP if response times warrant it.

## Changes

**Deleted**
- `src/data/entity/volunteer/volunteer-list-mv.entity.ts`
- `src/data/view/volunteer-list-mv.ts` (and the now-empty `data/view/` directory)
- `src/server/utils/data/serve-volunteer-list.ts`

**Stripped references** from
- `src/data/data-source.ts` (entity import + `entities` array)
- `src/server/plugins/typeorm.ts` (`volunteerListMvRepository` registration)
- `src/server/types/fastify.d.ts` (decorator type)
- `src/server/utils/data/index.ts` (re-export)

**No DB migration needed** — the view is already gone.

## Verified

- zero residual references to `VolunteerListMV` / `volunteerListMv` / `serve-volunteer-list` / `createVolunteerListMV`
- `yarn typecheck` clean
- `yarn test:run` — 203 tests pass
- `be` container reloads cleanly; health-check returns 200


need4deed (approved):
Approving: removes dead MV scaffolding pointing at an already-dropped view; tests + typecheck clean.

PR #580, opened 2026-05-27
refactor: abstract notify service over email + slack transports

arturasmckwcz:
## Summary
- Replace fastify-mailer-based email plugin with a transport-agnostic notify service under `src/services/notify/` (types, SES + Slack-webhook transports, `withFallback` dispatch helper, per-event modules). Swapping SES for SMTP/Resend later is a one-case change in `buildEmailTransport`.
- Slack webhook posts wired in for new opportunity and volunteer legacy submissions — uses the existing `SLACK_WEBHOOK_URL` env var; absent var degrades to no-op so the server still boots without Slack configured.
- Drop `fastify-mailer` and `nodemailer` deps; the SES transport calls `@aws-sdk/client-sesv2` directly.

## Notes
- `fastify.notify` decorator exposes `emailVerification(user)`, `opportunitySubmitted(title)`, `volunteerSubmitted(email)`.
- Verification email caller in `routes/user.ts` now logs failures explicitly (was silently swallowed inside the old plugin).
- Slack message format: `New opportunity arrived: "<title>" at YYYY-MM-DD HH:mm` (Europe/Berlin TZ, minute precision).
- Ops events absorb their own errors and log at warn level — a Slack outage will never break a form submission.

## Test plan
- [ ] `yarn install` (lockfile prune for removed deps)
- [ ] `yarn typecheck` ✅
- [ ] `yarn lint` ✅ (no new errors)
- [ ] `yarn test:run` ✅ (203 tests pass, ran via pre-push hook)
- [ ] Smoke: POST `/opportunity/legacy` with valid agent → 200 + Slack message lands
- [ ] Smoke: POST `/volunteer/legacy` → 200 + Slack message lands
- [ ] Smoke: POST `/user` → verification email still sends via SES
- [ ] Smoke: unset `SLACK_WEBHOOK_URL` and confirm submissions still succeed (slack events no-op)


need4deed (approved):
Approved.

Issue #590, opened 2026-05-29
Backfill: populate Opportunity.submittedByPersonId from existing rac_* comment blobs


arturasmckwcz:
## Context

Before #589 lands, every legacy submission stored the submitter's PII as a Comment row with `text = '<rac_email>|<|><rac_full_name><|><rac_address><|><rac_plz><|><rac_phone>'` (`src/server/routes/opportunity/legacy.routes.ts:94-102`). After #587 adds `Opportunity.submittedByPersonId` and #589 rewires the handler, only **new** submissions populate the FK. Historical opportunities still have their submitter info trapped in the blob.

This issue is a one-shot data migration that walks existing opportunities, parses the blob from their Comment row, and applies the *same* resolution logic as #589 to populate `submittedByPersonId`. **The Comment row is left in place untouched** — cleanup is intentionally out of scope here, since it's a separate decision about retention and may need legal/coord sign-off.

The codebase already supports data-only migrations under `src/data/migrations/` — the existing reference is `1779739039589-seed-matched-opp-vol.ts`. Use that style.

## Identifying the blob comment

The fingerprint of a legacy submission Comment is:

- `entityType = 'opportunity'` (`EntityTableName.OPPORTUNITY` in `node_modules/need4deed-sdk/dist/types/core.d.ts:63`)
- `entityId = <opportunity.id>`
- `userId = 1` (`legacy.routes.ts:100`)
- `text` contains **exactly four** `<|>` separators (5 fields: email, full name, address, plz, phone)

A robust filter:

```sql
WHERE entity_type = 'opportunity'
  AND user_id = 1
  AND text ~ '^[^|]*(<\|>[^|]*){4}$'
```

If multiple matching comments exist on one opportunity (shouldn't, but be defensive), take the **earliest** by `created_at` — that's the original submission. Skip opportunities that already have `submittedByPersonId` set (so the migration is idempotent and safe to re-run after partial application).

## Resolution logic (mirror #589 exactly)

For each (opportunity, parsed-blob) pair, with `agentId = opportunity.agentId`:

1. Look up `Person` by `rac_email` (case-insensitive). If `rac_email` is empty → log a warning and skip the row (do NOT create a nameless Person).
2. If found AND an `agent_person` row exists for `(person.id, agentId)` → set `opportunity.submittedByPersonId = person.id`.
[...]

Issue #588, opened 2026-05-29
ApiOpportunityGet.contact: prefer original submitter when still at agent


arturasmckwcz:
## Context

`getOpportunityContact` (`src/services/dto/dto-opportunity.ts:29-40`) currently derives `ApiOpportunityGet.contact` from `opportunity.agent.representative.person`, where `Agent.representative` is a getter that picks the first \`agentPerson\` row with role \`VOLUNTEER_COORDINATOR\`, falling back to \`agentPerson[0]\` (`src/data/entity/opportunity/agent.entity.ts:134-142`).

The consequence: the contact field can name a person who never submitted the opportunity — it's whoever the agent's coordinator happens to be at read time. That's a quiet identity substitution: an API consumer treating \`contact\` as 'who filed this' is wrong.

After #587 lands, \`Opportunity.submittedByPersonId\` becomes available. This issue makes \`contact\` prefer that person when they still hold an \`agent_person\` row for the opportunity's agent (i.e. they're still associated with the centre); otherwise it falls back to the existing coordinator/agent-member resolution.

\`ApiOpportunityGet\` lives in the SDK: \`node_modules/need4deed-sdk/dist/types/api/opportunity.d.ts:136-142\`, and the route response is shaped by \`dtoOpportunityGet\` (`src/services/dto/dto-opportunity.ts:120-173`), called from \`opportunity.routes.ts:144\`.

## Acceptance criteria

- \`getOpportunityContact(opportunity)\` returns the \`submittedByPerson\` when **both** are true:
  1. \`opportunity.submittedByPersonId\` is set
  2. There exists an \`agent_person\` row for \`(submittedByPersonId, opportunity.agentId)\`
- Otherwise falls back to the existing \`opportunity.agent.representative.person\` resolution unchanged.
- The route's TypeORM \`relations\` list at \`src/server/routes/opportunity/opportunity.routes.ts:81-91\` is extended so that the DTO has what it needs to make the decision (e.g. \`submittedByPerson\` and \`submittedByPerson.agentPerson\`, or load via a small repository call inside the DTO — pick whichever stays consistent with neighbouring DTOs).
- An SDK update or local JSDoc on \`ApiOpportunityContact\` documents: 'Best-effort: the original submitter if they are still listed at the opportunity's agent; otherwise the current agent representative. Not guaranteed to be the submitter.' Coordinate the SDK change per \`CLAUDE.md\`'s API contract section.
- Tests in \`src/test/services/dto/\` cover both branches:
[...]

PR #592, opened 2026-05-30
feat: add nullable submittedByPersonId FK to Opportunity (#587)

arturasmckwcz:
## Summary

- Adds `Opportunity.submittedByPersonId` (nullable) → `Person` and the matching `submittedByPerson` `@ManyToOne` relation, placed alongside the existing `agentId` block.
- Generated migration `1780167679271-add-submitted-by-person-to-opportunity.ts` adds the column + FK; existing rows default to `NULL`.

This is the structural-only change. No DTO, route, or seed wiring — those land in #588 (DTO), #589 (legacy handler), #590 (backfill).

Closes #587.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn test:run` — 203/203
- [x] Migration `up()`/`down()` are symmetric (add column + FK, drop FK + column)
- [ ] After merge: nodemon in the `be` dev container will pick up the migration on next restart per project convention

### DIFF
--- a/src/data/entity/opportunity/opportunity.entity.ts
+++ b/src/data/entity/opportunity/opportunity.entity.ts
@@ -18,6 +18,7 @@ import {
 import Deal from "../deal.entity";
 import District from "../location/district.entity";
 import OpportunityVolunteer from "../m2m/opportunity-volunteer";
+import Person from "../person.entity";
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Approved.

PR #595, opened 2026-05-30
feat: legacy POST resolves submitter, drops Comment PII blob (#589)

arturasmckwcz:
## Summary

Replaces the per-submission Comment row that legacy `POST /opportunity/legacy` used as a delimited-blob dumping ground for submitter PII (`rac_email<|>rac_full_name<|>rac_address<|>rac_plz<|>rac_phone`, written with `userId: 1`) with a structured Person + AgentPerson upsert. The link is recorded on `Opportunity.submittedByPersonId` (added in #587), which makes `ApiOpportunityGet.contact` (changed in #588) return the actual submitter when they're still at the agent.

### New helper: `getOrCreateSubmitterPerson`

Four-branch resolution (mirrors the issue's contract exactly):

1. Empty `rac_email` → return `null`, no Person manufactured.
2. Person found AND `agent_person` link exists for `(person, agentId)` → no-op.
3. Person found, no link → upsert `AgentPerson(role: VOLUNTEER_COORDINATOR)`.
4. Person not found → create Person from `rac_*` (firstName/lastName split from `rac_full_name`, email, phone) + upsert AgentPerson.

Signature is `(body, agentId, manager?: DataSource | EntityManager)` — same pattern as `createAddress` / `patchAddress` in `for-routes.ts`, so #590's backfill migration can call the helper from a `QueryRunner.manager` without a Fastify dependency.

### Preserved behaviour (per the issue's open-questions section)

- Domain-based authZ in the preHandler is **kept as-is**. Tightening it (full email match or verification challenge) is a separate authZ-scope follow-up.
- Empty/missing `rac_email` still passes the preHandler (the `email ILIKE '%@'` quirk is unchanged); the resolver simply returns `null` and `submittedByPersonId` stays unset, so today's "submission proceeds against orphanage / first agent" behaviour is preserved.

### Scoped out

- **Address creation** from `rac_address` + `rac_plz`. Mentioned in the issue body but `Person.addressId` is nullable and the centre address is already on `Agent`. Easy to layer on later via `createAddress` from `for-routes.ts` when there's a concrete need.
- **Backfill** of existing opportunities' Comment blobs — that's #590.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn test:run` — 215/215 (was 208, +7 new)
- [x] All four resolution branches covered, plus name-split edge cases (whitespace-only name falls back to email local-part; single-token name leaves `lastName` undefined; multi-word name joins trailing tokens).

## Files changed

- `src/server/utils/data/get-or-create-submitter-person.ts` — new helper
- `src/server/utils/data/index.ts` — barrel export
- `src/server/routes/opportunity/legacy.routes.ts` — call resolver, remove Comment side-write
- `src/test/server/utils/data/get-or-create-submitter-person.test.ts` — 7 tests

Closes #589.

### DIFF
--- a/src/server/routes/opportunity/legacy.routes.ts
+++ b/src/server/routes/opportunity/legacy.routes.ts
@@ -4,13 +4,11 @@ import {
   FastifyPluginOptions,
 } from "fastify";
 import {
-  EntityTableName,
   OpportunityLegacyFormData,
   OpportunityStatusType,
 } from "need4deed-sdk";
 import { ILike, In } from "typeorm";
 import { UnauthorizedError } from "../../../config";
-import Comment from "../../../data/entity/comment.entity";
 import Agent from "../../../data/entity/opportunity/agent.entity";
 import Opportunity from "../../../data/entity/opportunity/opportunity.entity";
 import logger from "../../../logger";
@@ -26,6 +24,7 @@ import {
   getDistrictToOpportunityHandler,
   getOpportunityNotificationText,
   getOpportunityOrphanageAgent,
+  getOrCreateSubmitterPerson,
   writeOpportunityLegacy,
 } from "../../utils";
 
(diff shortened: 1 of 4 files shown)


need4deed (approved):
Approved.

PR #596, opened 2026-05-30
feat: backfill Opportunity.submittedByPersonId from legacy comment blobs (#590)

### LINKED ISSUE #590: Backfill: populate Opportunity.submittedByPersonId from existing rac_* comment blobs

arturasmckwcz:
## Summary

One-shot data migration that catches up the historical opportunities created before #589 — they still have their submitter PII trapped in a `<|>`-delimited Comment row (`text = email<|>full_name<|>address<|>plz<|>phone`, `user_id = 1`, `entity_type = 'opportunity'`). The migration walks them, parses the blob, and populates the new `submitted_by_person_id` FK via `getOrCreateSubmitterPerson` — the same helper the runtime handler uses, so there's a single source of truth for the four-branch resolution.

### Identification

Blob rows are matched by the fingerprint described in #590:

```sql
entity_type = 'opportunity'
  AND user_id = 1
  AND text ~ '^[^|]*(<\|>[^|]*){4}$'  -- exactly 4 separators
```

Earliest matching comment per opportunity wins (`DISTINCT ON (o.id) ... ORDER BY c.created_at ASC`). Opportunities that already have `submitted_by_person_id` set are filtered out at the `WHERE` level, so the migration is naturally idempotent.

### Skip semantics

- **Empty `rac_email`** → log warning, skip. No nameless Person manufactured (matches runtime behaviour from #589).
- **Opportunity with no `agent_id`** → log warning, skip. The runtime fallback to orphanage was deliberately not extended here, since misattributing history feels worse than leaving the FK null.
- **Already set** → never enters the loop (SQL filter).

### Reversibility

`down()` nulls the column on all rows. Person/AgentPerson rows possibly created in `up()` are kept — they're independently valid (the person legitimately works at the agent) and can't be reliably identified after the fact.

### Verified on local dev DB

Found 1 legacy blob (opportunity 840), processed cleanly:

```
[backfill] summary: {"alreadySet":0,"eligible":1,"foundLinked":0,
                     "foundNoLink":0,"created":1,
                     "skippedNoEmail":0,"skippedNoAgent":0}
```

Person 1383 (`Good Samaritan <gs@moveglobal.de>`) created, AgentPerson link to agent 149 upserted, opportunity 840's `submitted_by_person_id` updated.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn test:run` — 215/215
- [x] `yarn migration:run` against local DB applies + commits cleanly
- [x] Per-branch counts logged via finally so summary surfaces even on abort
- [ ] After merge: run against pre/prod DBs. Resync sequences first if they're behind (see [DB sequence desync](https://github.com/.../memory) — also recently surfaced for #587's migration).

## Scoped out (per #590)

- Deleting/redacting the backfilled Comment rows — separate retention decision.
- Backfilling opportunities whose blob has been hand-edited and no longer matches the 4-separator regex.

## Notes for the deployer

This migration writes to `person` and `agent_person` (`person_id_seq`, `agent_person_id_seq`). Pre/prod sequences should be checked for desync before running — the resync recipe lives in our ops notes. The migration itself runs inside TypeORM's per-migration transaction, so a partial failure rolls back cleanly and leaves the migration unapplied for retry.

Closes #590.

### DIFF
--- a/src/data/migrations/1780169787517-backfill-opportunity-submitted-by-person.ts
+++ b/src/data/migrations/1780169787517-backfill-opportunity-submitted-by-person.ts
@@ -1,0 +1,121 @@
+import { ILike, MigrationInterface, QueryRunner } from "typeorm";
+import { getOrCreateSubmitterPerson } from "../../server/utils/data/get-or-create-submitter-person";
+import AgentPerson from "../entity/m2m/agent-person";
+import Person from "../entity/person.entity";
+
+// One-shot data migration: walk Opportunity rows that still carry a legacy
+// rac_* Comment blob (text shaped as `email<|>full_name<|>address<|>plz<|>phone`,
+// userId 1, entityType 'opportunity'), parse the blob, and populate
+// Opportunity.submitted_by_person_id via the same getOrCreateSubmitterPerson
+// helper the runtime handler uses (#589). The Comment row is left in place.
+//
+// Idempotent: only touches opportunities whose submitted_by_person_id is null.
+// Reversible: down() nulls the column on all rows; Person / AgentPerson rows
+// possibly created by up() are kept (they can't be reliably identified after
+// the fact and are safe to retain).
+
+interface BlobRow {
+  opportunity_id: number;
+  agent_id: number | null;
+  blob: string;
+}
+
+export class BackfillOpportunitySubmittedByPerson1780169787517
+  implements MigrationInterface
+{
+  name = "BackfillOpportunitySubmittedByPerson1780169787517";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const manager = queryRunner.manager;
+
+    const [alreadySet]: [{ count: string }] = await manager.query(
+      `SELECT COUNT(*)::text AS count FROM opportunity WHERE submitted_by_person_id IS NOT NULL`,
+    );
+
+    const rows: BlobRow[] = await manager.query(`
+      SELECT DISTINCT ON (o.id)
+        o.id AS opportunity_id,
+        o.agent_id AS agent_id,
+        c.text AS blob
+      FROM opportunity o
+      JOIN comment c
+        ON c.entity_type = 'opportunity'
+       AND c.entity_id = o.id
+       AND c.user_id = 1
+       AND c.text ~ '^[^|]*(<\\|>[^|]*){4}$'
+      WHERE o.submitted_by_person_id IS NULL
+      ORDER BY o.id, c.created_at ASC
+    `);
+
+    const counts = {
+      alreadySet: Number(alreadySet?.count ?? 0),
+      eligible: rows.length,
+      foundLinked: 0,
+      foundNoLink: 0,
+      created: 0,
+      skippedNoEmail: 0,
+      skippedNoAgent: 0,
+    };
+
+    try {
+      for (const row of rows) {
+        const oppId = row.opportunity_id;
+        if (!row.agent_id) {
+          counts.skippedNoAgent++;
+          console.warn(`[backfill] opp ${oppId}: no agent_id; skipping`);
+          continue;
+        }
+
+        const parts = row.blob.split("<|>");
+        const [rac_email, rac_full_name, , , rac_phone] = parts;
+
+        if (!rac_email?.trim()) {
+          counts.skippedNoEmail++;
+          console.warn(`[backfill] opp ${oppId}: empty rac_email; skipping`);
+          continue;
+        }
+
+        const preExistingPerson = await manager.findOne(Person, {
+          where: { email: ILike(rac_email.trim()) },
+        });
+        const preExistingLink = preExistingPerson
+          ? await manager.findOne(AgentPerson, {
+              where: { agentId: row.agent_id, personId: preExistingPerson.id },
+            })
+          : null;
+
+        const person = await getOrCreateSubmitterPerson(
+          { rac_email, rac_full_name, rac_phone },
+          row.agent_id,
+          manager,
+        );
+
+        if (!person) {
+          counts.skippedNoEmail++;
+          continue;
+        }
+
+        await manager.query(
+          `UPDATE opportunity SET submitted_by_person_id = $1 WHERE id = $2 AND submitted_by_person_id IS NULL`,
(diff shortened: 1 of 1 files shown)


need4deed (approved):
Approved.

PR #603, opened 2026-06-01
feat: GET /user — add role filter + honour sortOrder (#602)

arturasmckwcz:
## Summary

Unblocks FE PR [need4deed-org/fe#537](https://github.com/need4deed-org/fe/pull/537) (autocomplete `@`-mentions in comments). The FE calls `GET /user?role=coordinator` to populate the mention menu; today's `GET /user` ignores `role`, has default `limit=12`, and silently ignores `sortOrder` — so the FE would only see at most 12 coordinators (and possibly zero, if the most-recent dozen users are non-coordinators).

### Changes

- **`sdk-types.json`** — add `UserRole` (was inlined in `ApiUserMe`; now reusable as `$ref: UserRole#`).
- **`userListQuerySchema`** — add `role` (`$ref: UserRole#`) and `sortOrder` (existing `sortOrderProps`).
- **`QuerystringUserList`** — extend with `role?: UserRole` and `sortOrder?: SortOrder`.
- **`getUserWhere(search, role)`** — applies both clauses at the top level (AND semantics in TypeORM). Pure-function tested across all four combinations.
- **`GET /user` handler** — destructures `role` + `sortOrder` from the query; passes `role` to `getUserWhere`, picks `id ASC` for `OldToNew`, `DESC` otherwise. Existing callers (no params) see no behaviour change.

No auth change. `GET /user` remains accessible to any authed user, by design — the comment-tagging case it supports is the documented reason for that decision.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn test:run` — 219/219 (was 215, +4 new for `getUserWhere`)
- [x] Local server boots with the new schema (no startup-time route-registration error)
- [ ] After merge: FE consumes the new contract via `?role=coordinator`. Confirm pagination is sufficient (default `limit=12`, max `120`) once the FE switches to filtered call.

## Out of scope

- Raising `paginationProps.limit.maximum` (120) — coordinator counts well below that today.
- Switching the order column from `id` to `createdAt` — kept on `id` for safety; monotonically increasing so equivalent ordering.
- The `ApiUserMe.role` schema still inlines the enum (line 326-328 of `sdk-types.json`); could be flipped to `$ref: UserRole#` in a follow-up cleanup. Not strictly needed for this PR's contract.

Closes #602.

### DIFF
--- a/src/server/routes/user.ts
+++ b/src/server/routes/user.ts
@@ -1,6 +1,6 @@
 import { validate } from "class-validator";
 import { FastifyInstance, FastifyPluginOptions } from "fastify";
-import { ApiUserGet, UserRole } from "need4deed-sdk";
+import { ApiUserGet, SortOrder, UserRole } from "need4deed-sdk";
 import { FindOptionsWhere } from "typeorm";
 import Person, { PersonUpdateType } from "../../data/entity/person.entity";
 import User from "../../data/entity/user.entity";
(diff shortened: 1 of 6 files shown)


need4deed (approved):
Approved.

PR #607, opened 2026-06-02
feat: comment-person m2m for @tagging in comments (fe#537)

arturasmckwcz:
## Summary

BE plumbing for FE PR [need4deed-org/fe#537](https://github.com/need4deed-org/fe/pull/537) (autocomplete `@`-mentions in comments). The SDK already declares `ApiComment.taggedPersonIds: number[]` as of `need4deed-sdk@0.0.97` (upgraded in this PR); this lands the storage + handler-side support.

### Schema

- **New m2m entity `CommentPerson(commentId, personId)`** — cascades on both sides, **plus a `@Unique(commentId, personId)` index** so duplicate tag rows can't slip in via races or direct INSERT. Generated migration adds the table, FKs, and the unique index.
- **Reverse `OneToMany` relations** on `Comment` and `Person`.
- **Request body** (`entity-types.json` `Comment#`) now accepts `taggedPersonIds: integer[]`.
- **Response shape** (`sdk-types.json` `ApiComment#`) exposes `taggedPersonIds: integer[]`.

### Helper

`syncCommentTags(commentId, desiredPersonIds, manager)` does the diff — delete removed rows, insert new ones, leave unchanged rows alone. Idempotent. Accepts a transactional `EntityManager` so callers can wrap it. Same pattern as `updateAgentLanguages` in `for-routes.ts`. JSDoc explicitly notes `undefined -> clear-all` semantic and points to the PATCH handler as the example of "guard at the call site for leave-untouched" semantics. 6 pure-function tests cover insert+delete diff, no-op, full-clear, undefined-as-clear, dedupe, all-new.

### Routes

- **`POST /comment`** — destructures `taggedPersonIds` out of the body, saves Comment + `CommentPerson` rows inside one transaction, returns the serialized `ApiComment` with tags populated.
- **`PATCH /comment/:id`** — same flow; the tag-sync only fires when `taggedPersonIds` is present in the patch body (PATCH semantics: omitting the field leaves existing tags alone).
- Both classify Postgres `23503` on the `comment_person` table as `400 "Invalid tagged person id"` — matched on `error.table === "comment_person"`, not detail-text substring, so it can't misfire on unrelated `person`-named constraints later.
- Response schema bumped from `Comment#` (entity) to `ApiComment#` on POST/PATCH so the FE gets a consistent payload. **Verified safe**: the only known consumer (`useCreateComment` / `useUpdateComment` in fe#537) reads only `{ message }` and re-fetches via the invalidated query key.
- `GET /comment` loads `commentPerson`; `addComments2Entity` (used by opportunity/agent/volunteer detail pages) does too — so existing list/detail responses pick up the new field automatically.

### DTO

`commentSerializer` adds `taggedPersonIds: comment.commentPerson?.map(cp => cp.personId).filter(Boolean) ?? []` — empty array when the relation isn't loaded or no tags exist.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn test:run` — **227/227** (was 219, +8 new: 6 `syncCommentTags`, 2 `dto-comment` extensions)
- [x] Migration applied on dev DB; `comment_person` table + unique index + FKs verified via `\d`
- [x] FE-side grep confirms POST/PATCH response-shape change is safe (only `{message}` is consumed)
- [ ] After merge: FE PR fe#537 exercises the round-trip; confirm POST/PATCH/GET return `taggedPersonIds` correctly.

## Cross-PR caveat for fe#537

The FE PR currently stores `<@user.id>` markers (user id), but the SDK + this PR expect `taggedPersonIds` (person id), and `ApiUserGet` doesn't expose `personId`. FE will need either (a) switch the marker to person id + extend `ApiUserGet` with `personId`, or (b) have the BE translate user->person on its end. Not addressed here.

## Out of scope (acknowledged follow-ups)

- Pre-validate `personIds` upfront instead of relying on PG FK violation — cheap-enough today, revisit if batched tagging becomes a thing.
- Add the same `@Unique` constraint retroactively to `AgentPerson` — separate ticket; not introduced or worsened by this PR.

Closes the BE half of fe#537 (issue need4deed-org/fe#503).

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.94",
+    "need4deed-sdk": "0.0.97",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
     "pino-pretty": "^13.0.0",
--- a/src/data/data-source.ts
+++ b/src/data/data-source.ts
@@ -18,6 +18,7 @@ import Postcode from "./entity/location/postcode.entity";
 import AgentLanguage from "./entity/m2m/agent-language";
 import AgentPerson from "./entity/m2m/agent-person";
 import AgentPostcode from "./entity/m2m/agent-postcode";
+import CommentPerson from "./entity/m2m/comment-person";
 import DistrictPostcode from "./entity/m2m/district-postcode";
 import LocationAddress from "./entity/m2m/location-address";
 import LocationDistrict from "./entity/m2m/location-district";
@@ -70,6 +71,7 @@ export const dataSource = new DataSource({
     Appreciation,
     Category,
     Comment,
+    CommentPerson,
     Communication,
     Config,
     Deal,
--- a/src/data/entity/comment.entity.ts
+++ b/src/data/entity/comment.entity.ts
@@ -6,8 +6,10 @@ import {
   Index,
   JoinColumn,
   ManyToOne,
+  OneToMany,
   PrimaryGeneratedColumn,
 } from "typeorm";
+import CommentPerson from "./m2m/comment-person";
 import Language from "./profile/language.entity";
 import User from "./user.entity";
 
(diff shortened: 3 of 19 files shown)


need4deed (approved):
Approved.

PR #610, opened 2026-06-02
feat: GET /comment supports taggedPersonId filter (#609)

arturasmckwcz:
## Summary

Adds the reverse-lookup that PR #607 left open: *"give me every comment where person X is tagged."* New optional `taggedPersonId` query param on `GET /comment`, composable with the existing `userId` (comment **author**), `entityId`, and `entityType` filters.

Closes #609.

## What it does

- **`GET /comment?taggedPersonId=7`** → returns every comment where person 7 appears in `comment_person`.
- **Composable** — `?taggedPersonId=7&userId=12&entityType=volunteer&entityId=42` narrows to "comments authored by user 12 on volunteer 42 where person 7 is tagged."
- **Person id, not user id** — matches the m2m column and the SDK's `taggedPersonIds: number[]`. No `userId` alias for tag lookups (would re-introduce the user/person ambiguity that fe#537 is still untangling).

## Two-step lookup (why not relation filter)

The natural `where: { commentPerson: { personId } }` approach has a subtle bug: TypeORM's relation filter *also* constrains the relation load, so `comment.commentPerson` in the result only contains the matching tag — the serializer would then report `taggedPersonIds: [X]` and hide every other tag on the comment.

Two-step query keeps the DTO honest:

1. Find `comment_person.comment_id` where `person_id = taggedPersonId`.
2. Load comments with `id IN (...)` and full relations.

Empty match short-circuits with a 200 + `data: []` before the second query (also avoids the `id: In([])` antipattern).

## Index

Adds `@Index(["personId"])` on `CommentPerson` (non-unique). The composite `(comment_id, person_id)` unique from #607 can only seek on `comment_id` first; this gives the reverse lookup its own seek path. Migration `add-index-comment-person-person-id` generated cleanly.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn test:run` — **229/229**
- [x] Migration applied locally; `comment_person` shows the new btree on `person_id` in `\d`
- [ ] After merge: the FE side can call `GET /comment?taggedPersonId=N` once it has a person id to query with (cross-PR caveat from fe#537 still applies — `ApiUserGet` doesn't yet expose `personId`).

## Out of scope

- No handler-level integration test added; consistent with the existing comment route which has no route-test coverage. Helper-level coverage (`syncCommentTags`) remains.
- No backfill — this is read-side only.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/data/entity/m2m/comment-person.ts
+++ b/src/data/entity/m2m/comment-person.ts
@@ -11,6 +11,9 @@ import Person from "../person.entity";
 
 @Entity()
 @Index(["commentId", "personId"], { unique: true })
+// Reverse-lookup index: "give me every comment where this person is tagged."
+// The composite unique above can seek on commentId-first only.
+@Index(["personId"])
 export default class CommentPerson {
   constructor(commentPerson?: Partial<CommentPerson>) {
     if (commentPerson) {
--- a/src/data/migrations/1780432006040-add-index-comment-person-person-id.ts
+++ b/src/data/migrations/1780432006040-add-index-comment-person-person-id.ts
@@ -1,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddIndexCommentPersonPersonId1780432006040
+  implements MigrationInterface
+{
+  name = "AddIndexCommentPersonPersonId1780432006040";
+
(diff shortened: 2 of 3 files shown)


need4deed (approved):
Approved.

PR #621, opened 2026-06-03
refactor: Deal ↔ Language m2m + drop Profile entity

arturasmckwcz:
Part of #611 (Phase 2). Closes #614. **Completes the Profile flattening** — after this, `Deal` joins activities, skills and languages directly and the `Profile` graph is gone.

## Changes

**Schema**
- New `DealLanguage` entity (carries `proficiency` + `purpose`, mirroring `ProfileLanguage`) with inverse relations on `Deal`/`Language`. **No unique constraint** — unlike activity/skill, a deal can legitimately hold the same language more than once (24 such `(deal, language)` pairs exist in the data).
- `info` and `category` (`category_id` FK) move from `Profile` onto `Deal`.
- `Profile` and all `Profile<->X` join entities/tables are dropped (`profile`, `profile_activity`, `profile_skill`, `profile_language`). `Category`'s inverse relation now points at `Deal`.

**Code**
- Every parser / write helper / DTO / where-builder / relation string / translation loop switched `profile.profileLanguage → deal.dealLanguage` and `deal.profile.{categoryId,info} → deal.{categoryId,info}`.
- `getCategoryToProfileHandler` → `getCategoryToDealHandler` (derives category from `deal.dealActivity`, stores on the deal); `fastify.db.profileRepository` → `dealRepository`.
- Removed the dead `AccompanyingLanguage` entry from the m2m relation map (no entity, no caller) and trimmed now-unused host-union members.

**Migration** `flatten-deal-language-drop-profile`
- Creates `deal_language` (+ enum types) and the `deal.info`/`category_id` columns, **backfills** `deal_language` and `deal.{info,category_id}` from the profile graph **before** dropping `deal.profile_id`, then drops the whole profile graph. Enum values are cast via text since the `deal_language_*` enum types are distinct from the `profile_language_*` ones. `down()` restores the structure (data not recovered).

## Verification
- Migration applied: `deal_language` = **4233** rows (exact match to deal-attached `profile_language`); `deal` info/category backfilled (281/387 rows); all `profile*` tables dropped.
- `yarn typecheck` ✓ · `yarn test:run` → 229 passed ✓ · lint clean for changed files.
- API response shapes unchanged.

## Heads-up for reviewers with a local dev DB
If you manually created the experimental `volunteer_list_mv` materialized view (from the #606/#608 perf spike), it depends on `deal.profile_id` and will block this migration locally — `DROP MATERIALIZED VIEW IF EXISTS volunteer_list_mv;` first. No committed migration creates that view (it was dropped by `1763036250587`), so CI/prod are unaffected.


need4deed (approved):
Approved: Deal ↔ Language m2m + Profile teardown. Migration backfills before dropping profile graph; tests green.

PR #622, opened 2026-06-03
chore: remove dead refreshMaterializedView util (stray volunteer_list_mv)

arturasmckwcz:
Removes the dead code left behind by the `volunteer_list_mv` materialized-view experiment (#606/#608 perf spike).

## Why
`volunteer_list_mv` is a **stray**: no committed migration creates it (migration `1763036250587` actually drops it), and the `refreshMaterializedView` helper that targeted it is **never called** — only re-exported from the data-utils barrel. It's dead weight, and the view itself depends on `deal.profile_id`, which blocks the Deal-flattening migrations.

## Changes
- Delete `src/data/utils/refresh-materialized-view.ts` and its barrel export.

## Not in this diff (git-ignored, local only)
The view physically survives only in the git-ignored local dev dumps (`data-bootstrap/dev.dump.sql`, `dev/business-data/*.sql`). I stripped the three `volunteer_list_mv` pg_dump sections from those locally so a fresh bootstrap no longer recreates it — but they're ignored, so nothing to commit. Heads-up for other devs: if your local DB has the view (from an old dump), `DROP MATERIALIZED VIEW IF EXISTS volunteer_list_mv;` to match.

## Verification
`yarn typecheck` ✓ · `yarn test` → 229 passed ✓ · lint clean.


need4deed (approved):
Approved: remove dead refreshMaterializedView util / stray MV cleanup.

PR #627, opened 2026-06-03
chore: upgrade need4deed-sdk to 0.0.98

arturasmckwcz:
Upgrade `need4deed-sdk` 0.0.97 → 0.0.98.

0.0.98 removes the leftover `new_api.ts` (need4deed-org/sdk#121) that wrongly modeled `Deal` as part of the API contract. That file also held the `DealType` enum, which a few BE files were importing **from the SDK** — so the bump alone broke `yarn typecheck`.

`DealType` is internal (Deal is storage, not contract), so this repoints those imports to the **local** `data/types` enum that `deal.entity.ts` already uses for the `deal_type` DB column:
- `services/dto/parser-deal-volunteer.ts`
- `services/dto/parser-deal-opportunity.ts`
- `data/seeds/types.ts`

No behavior change — same enum values (`volunteer`/`opportunity`).

## Verification
`yarn typecheck` ✓ · `yarn test:run` → 229 passed ✓ · lint clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.97",
+    "need4deed-sdk": "0.0.98",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
     "pino-pretty": "^13.0.0",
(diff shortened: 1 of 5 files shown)


need4deed (approved):
Approved: upgrade need4deed-sdk to 0.0.98; DealType moved to local internal enum. Tests green.

PR #632, opened 2026-06-04
fix(opportunity): refresh submitter Person fields from rac_* on resubmit

arturasmckwcz:
## Problem

`POST /opportunity` resolves the submitter via `getOrCreateSubmitterPerson(body, agentId)`, which looks the Person up by `rac_email`. When the Person **already exists**, the helper left its fields untouched. So a record with a blank `first_name` (or missing phone) was never corrected on resubmission — and `ApiOpportunityGet.contact` surfaced an empty `name` even though the form carried a real `rac_full_name` (observed on prod opp #876, contact person id 63).

## Change

When the Person exists, overwrite the `rac_*`-backed fields from the current submission:

- `rac_full_name` → `firstName` / `lastName`
- `rac_phone` → `phone`

**Selective by design:** a field is written only when the form actually provides a value, so a later submission that omits `rac_full_name` / `rac_phone` cannot wipe an existing good value (i.e. it can't reintroduce the empty-name bug in reverse). `rac_email` remains the identity key and is not overwritten. A `dirty` flag skips the save when nothing changed.

The `AgentPerson` link upsert is unchanged and still covers both the found and created cases (an existing person submitting for a new agent still gets linked).

## Requirements covered

1. Identify person by `rac_email` — existing
2. Not found → create + link to the figured agent — existing
3. **Found → overwrite corresponding `rac_*` fields — added**

## Tests

`get-or-create-submitter-person.test.ts`: updated branches 2 & 3 to assert the refresh, added a "no incoming values → record untouched" case and a selective-overwrite (phone-only, name preserved) case. 9/9 pass; `yarn typecheck` clean.

### DIFF
--- a/src/server/utils/data/get-or-create-submitter-person.ts
+++ b/src/server/utils/data/get-or-create-submitter-person.ts
@@ -79,11 +79,15 @@ export async function getOrCreateSubmitterPerson(
     if (fullName) {
       const { firstName, lastName } = splitName(fullName, email);
       person.firstName = firstName;
-      person.lastName = lastName;
+      // Use null (not undefined) when the new name has no last token: TypeORM
+      // skips undefined on update, which would leave a stale last name behind
+      // (e.g. resubmitting "Cher" over "Mary van der Berg" -> "Cher van der Berg").
+      person.lastName = lastName ?? (null as unknown as undefined);
       dirty = true;
     }
-    if (body.rac_phone) {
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Approved: refreshes submitter Person fields (name/phone) on resubmit; selective overwrite avoids clobbering, lastName cleared correctly.

PR #633, opened 2026-06-04
feat(opportunity): patch submitter Person address from rac_address / rac_plz

arturasmckwcz:
Extends `getOrCreateSubmitterPerson` (POST /opportunity) to also resolve the submitter's **Address**, alongside the name/phone refresh already in place.

## Behaviour

- **`rac_address` → `address.street`** — strips a trailing German 5-digit postcode and the city after it (`streetFromAddress`), e.g. `"Musterstr. 1, 12345 Berlin"` → `"Musterstr. 1"`.
- **`rac_plz` → an *existing* `Postcode`** (looked up by `value`, never created):
  - existing address + `rac_plz` resolves → update the postcode
  - existing address + `rac_plz` unknown → **leave the postcode untouched**
  - no address yet → **create** one, falling back to **`12345`** when `rac_plz` doesn't resolve (an `Address` requires a `Postcode` — NOT NULL)

## Notes / decisions

- Reuses the existing `patchAddress` / `createAddress` helpers in `for-routes.ts` (per the "reuse for-routes helpers" convention) — `patchAddress` already resolves postcode by value and leaves it untouched when unresolved, which matches the "figure existing / leave as is" requirement.
- The fallback-to-`12345` applies **only when creating a brand-new address**; for an existing address an unknown `rac_plz` leaves the current postcode in place. This split is forced by the `Address.postcode_id` NOT NULL constraint and resolves the "leave as is *or* fallback" ambiguity. ⚠️ Assumes a `Postcode` row with `value = '12345'` exists (the seeded dummy); if it doesn't resolve, no address is created and the submitter is still saved — no crash.
- `rac_address` / `rac_plz` are **optional** on `SubmitterFields`, so the existing backfill migrations (which only carry the name/phone/email blob) still type-check and skip the address step.

## Tests

`get-or-create-submitter-person.test.ts`: added an `address` group (patch existing, unknown-plz leaves postcode, create+link for new person, fallback-to-12345, no-op when neither field present) and a parametrised `streetFromAddress` group. 22/22 pass; `yarn typecheck` clean.

### DIFF
--- a/src/server/utils/data/get-or-create-submitter-person.ts
+++ b/src/server/utils/data/get-or-create-submitter-person.ts
@@ -22,6 +22,11 @@ type SubmitterFields = Pick<
 // fall back to this value rather than dropping the address entirely.
 const FALLBACK_PLZ = "12345";
 
+// Seeded placeholder address (see seeds/user.seed.ts) used for "unknown
+// address". It is shared by many Person rows, so it must never be patched in
+// place — a fresh Address is minted instead.
+const DUMMY_ADDRESS_TITLE = "Dummy";
+
 /**
  * Extract the street portion of a free-text address, cutting the German 5-digit
  * postcode and the city that follows it (e.g. "Musterstr. 1, 12345 Berlin" ->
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Approved: submitter address (rac_address/rac_plz) + getNameFields split; shared Dummy address protected via fresh-address-and-repoint. Tests pass.

PR #634, opened 2026-06-04
perf(volunteer): two-step list query + listType-driven relations

arturasmckwcz:
## Problem

`GET /volunteer` took **~18-20s on every call** in production (half a day after deploy, and reproduced from a second client against the already-warm DB — so **not** a cache/snapshot-lazy-load effect; the query is genuinely expensive every time).

**Cause:** the handler loaded five one-to-many `deal_*` collections (`dealLanguage`, `dealDistrict`, `dealActivity`, `dealSkill`, `dealTimeslot`) in a single `findAndCount`. So the **COUNT ran over a 5-way cartesian join across the whole `volunteer` table** (unbounded by page size), and the page itself hydrated a cartesian product.

## Changes (backend only)

1. **Two-step pagination**
   - *Step 1* — page volunteer **ids + count** with no collection hydration. TypeORM auto-joins only the relations the `where` filter actually touches, so the COUNT is cheap and the filter joins become **dynamic** (none when unfiltered).
   - *Step 2* — load the page's entities via `relationLoadStrategy: "query"` (one query per relation, **no cartesian**).
2. **`listType` query param** (`"card" | "table"`, default `"card"`) selects the relation set, matching what each FE view renders:
   - `table` → `person` + `languages` + `locations`
   - `card` → the above + `activities` + `skills` + `availability`
   - `person.address.postcode` and `deal.postcode` dropped from the list entirely (the list serializer never used them).
3. `volunteerListSerializer` defaults absent collections to `[]` so the response shape stays stable.

## Contract / SDK

`QuerystringVolunteerGetList` is local to `be` and the response shape is unchanged (table returns `[]` for the collections it doesn't render), so **no SDK change**. The request param is added to `volunteerListQuerySchema` + the local type.

## ⚠️ Deploy ordering

`volunteerListQuerySchema` has `additionalProperties: false`, so a FE sending `listType` to an **old** backend would 400. **Merge/deploy this BEFORE the FE PR** (FleetView fe: pass `listType`). `listType` defaults to `card`, so this backend is fully backward-compatible with the current FE.

## Verification

`yarn typecheck` clean, `dto-volunteer` tests pass, lint clean on touched files. Recommend a local `EXPLAIN (ANALYZE, BUFFERS)` before/after on the count query to quantify (the count no longer joins the collections).

### DIFF
--- a/src/server/routes/volunteer/volunteer.routes.ts
+++ b/src/server/routes/volunteer/volunteer.routes.ts
@@ -9,7 +9,7 @@ import {
   VolunteerFormData,
   VolunteerPatchBodyData,
 } from "need4deed-sdk";
-import { FindOptionsWhere } from "typeorm";
+import { FindOptionsOrder, FindOptionsWhere, In } from "typeorm";
 import DealActivity from "../../../data/entity/m2m/deal-activity";
 import DealDistrict from "../../../data/entity/m2m/deal-district";
 import DealLanguage from "../../../data/entity/m2m/deal-language";
@@ -30,7 +30,11 @@ import {
   responseSchema,
   volunteerListQuerySchema,
 } from "../../schema";
-import { QuerystringVolunteerGetList, RoutePrefix } from "../../types";
+import {
+  QuerystringVolunteerGetList,
+  RoutePrefix,
+  VolunteerListType,
+} from "../../types";
 import {
   fetchVolunteerById,
   getLanguageCode,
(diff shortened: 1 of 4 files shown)


need4deed (approved):
Approved: two-step volunteer list query + listType-driven relations. Verified locally — count no longer joins collections (302k-row cartesian + temp spill -> index-only scan); two-step executes correctly and listType trims relations.

PR #635, opened 2026-06-05
fix(user): serve personId from GET /user/me

arturasmckwcz:
## Problem

`ApiUserGet` (SDK contract) declares `personId`, but `GET /user/me` never returned it:
- `serializeUserToMeDTO` didn't set `personId`, and
- the `ApiUserMe` response schema (`sdk-types.json`) didn't list it, so Fastify's serializer stripped it even when present.

(The `GET /user` list shares the same DTO + schema, so it was missing `personId` too.)

## Fix

- Add `personId: user.personId` to `serializeUserToMeDTO` (`personId` is a plain column on `User`, already loaded — no extra relation/query).
- Add `"personId": { "type": "number" }` to the `ApiUserMe` response schema.

`personId` is voidable in the contract, so it's not added to `required`. **No SDK change** — the contract already declares it; the backend just wasn't honouring it.

## Tests

Updated `dto-user.test.ts` to assert `personId` maps through. `yarn typecheck` clean, 3/3 DTO tests pass, schema JSON validated.

### DIFF
--- a/src/server/schema/sdk-types.json
+++ b/src/server/schema/sdk-types.json
@@ -321,6 +321,9 @@
         "id": {
           "type": "number"
         },
+        "personId": {
+          "type": "number"
+        },
         "email": {
           "type": "string"
         },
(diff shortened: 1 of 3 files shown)


need4deed (approved):
Approved: serve personId from /user/me + GET /user; null personId now omitted (no spurious 0). Tests pass.

PR #641, opened 2026-06-08
feat(opportunity): back up RAC contact as a piped comment on submission

arturasmckwcz:
## What

On `POST /opportunity` (public form), persist the submitter's RAC contact as a piped `<|>` comment — **in addition to** the existing Person-based flow, not instead of it.

- `getOrCreateSubmitterPerson` → `submittedByPerson` → `getOpportunityContact` → `ApiOpportunityGet.contact` stays the primary source (unchanged).
- New `writeOpportunityContactComment` also stores `email<|>name<|>address<|>plz<|>phone` as a durable backup, so submitted contact data can't be lost if something goes wrong with the Person path — and it's the fallback the dashboard already reads for opportunities without `.contact`.

## Why

Belt-and-suspenders for contact data (re: #636). The current FE reads `.contact` primarily and piped comments only as a legacy fallback, so this is a safety net, not the display path.

## Design

- **Additive** — the submitter-setting workflow is untouched.
- **Best-effort** — `writeOpportunityContactComment` swallows its own errors and no-ops when no contact fields are present, so it can never block or fail a submission.
- Author: a User linked to the agent if found, else any user (author isn't shown for piped comments; just satisfies `comment.user_id` NOT NULL).

## Tests

New `write-opportunity-contact-comment.test.ts`: piped format + agent-linked author, fallback author, no-op when empty, no-op when no user, and the swallow-on-failure guarantee. 5/5 pass; typecheck + lint clean.

## Relationship to #637
This reimplements #637's useful half (the submission-time contact write) cleanly, without #637's `volunteer.routes.ts` change (which conflicts with the merged #634 two-step) and without the comment-PATCH ownership relaxation (moot — the FE edits contact via `PATCH /opportunity/:id`, not the comment). **#637 can be closed in favour of this.**

### DIFF
--- a/src/server/routes/opportunity/legacy.routes.ts
+++ b/src/server/routes/opportunity/legacy.routes.ts
@@ -25,6 +25,7 @@ import {
   getOpportunityNotificationText,
   getOpportunityOrphanageAgent,
   getOrCreateSubmitterPerson,
+  writeOpportunityContactComment,
   writeOpportunityLegacy,
 } from "../../utils";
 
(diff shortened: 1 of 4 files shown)


need4deed (approved):
Approved: additive contact backup as a piped comment; submitter (.contact) flow unchanged; best-effort write can't block submission. Tests pass.

PR #643, opened 2026-06-08
fix(utils): make getStartEndDates validation guard actually validate

arturasmckwcz:
Surfaced while reviewing the data-utils tests in #528.

`getStartEndDates`'s guard was unsatisfiable:
```ts
if (startHour < endHour && startHour < 0 && startHour > 24 && endHour < 0 && endHour > 24)
```
`startHour < 0 && startHour > 24` can never both hold, so invalid hours never threw. Replaced with the intended check — throw unless `0 <= startHour < endHour <= 24`.

**Scope:** source-only. `getStartEndDates` has **no production callers today** (only the unit test in #528), so this is a correctness/hygiene fix with no runtime behavior change for live code, and the happy-path inputs in #528 (start < end, in range) still pass. The throw-path test belongs with the suite in #528, so this PR deliberately doesn't touch the test file (avoids a conflict with #528).

**Not included:** `getStartEnd`'s map has questionable entries (e.g. `"17:00-20:00" → 14..17`), but it has 4 real callers and some entries look like intentional slot-snapping — correcting it would change live availability parsing, so it needs domain review, not a blind fix. Flagged on #528.

### DIFF
--- a/src/data/utils/getStartEndDates.ts
+++ b/src/data/utils/getStartEndDates.ts
@@ -1,26 +1,3 @@
-export function getStartEndDates(
-  startHour: number,
-  endHour: number,
-  date?: Date,
-): { start: Date; end: Date } {
-  if (
-    startHour < endHour &&
-    startHour < 0 &&
-    startHour > 24 &&
-    endHour < 0 &&
-    endHour > 24
-  ) {
-    throw new Error("Invalid start or end hours");
-  }
-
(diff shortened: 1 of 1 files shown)


need4deed (approved):
Approved: remove stray getStartEndDates (no production callers); getStartEnd retained. Source-only, typecheck green.

PR #644, opened 2026-06-08
feat(auth): add POST /auth/logout (#581)

arturasmckwcz:
Closes #581.

## What

New **`POST /auth/logout`**:
1. **Endpoint** at `/auth/logout` (new `RoutePrefix.LOGOUT`).
2. **Authenticated** — `onRequest: [fastify.authenticate()]` (any valid session).
3. **Clears the httpOnly cookies** — `reply.clearCookie` for both `access` and `refresh`, using the same `cookieOptions` (path/sameSite/secure) they were set with so the browser actually removes them. Returns `{ message: "Logout successful." }`.

Also adds `jwt-auth-plugin` to the auth-routes `fp` dependencies so `fastify.authenticate` is guaranteed available.

## Notes

- `typecheck` clean; lint clean on the change (the 2 `any` warnings in this file are pre-existing in the login/refresh `Reply` types).
- No automated test: the repo has no auth-route test harness (login/refresh are untested too), and these routes are DB-backed + need a real JWT. Happy to add an `app.inject()`-based test if you'd like to start that harness.
- Design choice: logout requires auth (per the issue's "do auth"). If you'd prefer it to also clear cookies for an expired/invalid session (so a stale tab can always log out), I can drop the `authenticate` guard — say the word.

### DIFF
--- a/src/server/routes/auth.ts
+++ b/src/server/routes/auth.ts
@@ -214,7 +214,6 @@ async function authRoutes(
   fastify.post<{ Reply: { message: string } }>(
     prefixedPath + RoutePrefix.LOGOUT,
     {
-      onRequest: [fastify.authenticate()],
       schema: {
         response: {
           200: {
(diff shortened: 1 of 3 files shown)


need4deed (approved):
Approved: POST /auth/logout — open endpoint clears access/refresh httpOnly cookies (matching set options); inject test covers it. Open-logout/CSRF trade-off is intentional and low severity.

PR #647, opened 2026-06-08
refactor(user): POST /user uses ApiUserPost; UserRole/Lang schema; fewer try/catch

arturasmckwcz:
## What

Refactor `POST /user` registration per the SDK contract.

- **Body typed as `ApiUserPost`** (SDK bumped 0.0.99 → **0.0.101**). Drops the bespoke inline body type (and the vestigial `resolvedPerson` field — that's a `FastifyRequest` decoration).
- **`createUserBodySchema` matches `ApiUserPost`:**
  - `role` → `$ref "UserRole#"` (was inline `["user","admin"]`)
  - `language` → `$ref "Lang#"` — a **new `"en"|"de"` schema** added to `sdk-types.json`
  - `password`, `role`, `language` now **required**
  - dropped `isActive` (server-controlled — forced `false`) and `timezone` (uses the entity default `"CET"`)
- **Fewer try/catch** (per the "Handling throws" convention):
  - preHandler person-lookup no longer wraps `findOneBy` in try/catch — DB errors propagate to the global handler (same 500, less code).
  - duplicate email surfaced as **409 up front** via a pre-check throwing the existing `ConflictError` (`error/typeorm.ts`), instead of catching the Postgres `23505`. The DB unique constraint remains the backstop for the rare race.

## ⚠️ Contract tightening (FE coordination)

The body is now strictly `ApiUserPost`: **`role`, `language`, `password` are required**, and **`timezone`/`isActive` are rejected** (`additionalProperties: false`). The registration FE must send the `ApiUserPost` shape (it should, once on SDK 0.0.101). `language` drives the verification-email locale (#645), so sending `"de"` matters for German users.

Two minor behaviour notes:
- Duplicate-email response message is now the generic `ConflictError` message ("Conflict: Resource already exists") rather than the previous specific string — still **409**.
- A concurrent duplicate that slips past the pre-check hits the unique constraint → currently a 500 (rare race).

## Verification

`yarn typecheck` clean, lint clean (one pre-existing `any` warning in the `Reply` type), and the affected DTO + server-boot tests pass (the boot test validates the new `Lang#`/`UserRole#` `$ref`s resolve). The `communication` field churn from the SDK round-trip nets to no change vs `develop`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.99",
+    "need4deed-sdk": "0.0.101",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
     "pino-pretty": "^13.0.0",
--- a/src/server/routes/user.ts
+++ b/src/server/routes/user.ts
@@ -1,8 +1,9 @@
 import { validate } from "class-validator";
 import { FastifyInstance, FastifyPluginOptions } from "fastify";
-import { ApiUserGet, SortOrder, UserRole } from "need4deed-sdk";
+import { ApiUserGet, ApiUserPost, SortOrder } from "need4deed-sdk";
 import { FindOptionsWhere } from "typeorm";
-import Person, { PersonUpdateType } from "../../data/entity/person.entity";
+import { ConflictError } from "../../config";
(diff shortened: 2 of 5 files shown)


need4deed (approved):
Approved: POST /user uses ApiUserPost + UserRole/Lang schema; throw-based handler; role-gated registration (admin/coordinator 403, agent email-domain gate, volunteer/user open).

PR #648, opened 2026-06-08
fix(user): default language to "en" when omitted on POST /user

arturasmckwcz:
SDK **0.0.102** makes `ApiUserPost.language` optional, so registration must default a missing `language` to `"en"`.

- **SDK** bumped 0.0.101 → 0.0.102.
- **Schema** (`createUserBodySchema`): `language` dropped from `required`; now `language: { allOf: [{ $ref: "Lang#" }], default: "en" }` — `allOf` keeps the `Lang` enum (en/de) while allowing the sibling `default`, which a bare `$ref` would ignore under draft-07.
- **Handler**: `language: language ?? Lang.EN` as a backstop (and the `User` entity column also defaults to `"en"`), so an omitted language always lands on `"en"`.

`yarn typecheck` + lint clean (one pre-existing `any` warning); the server-boot test passes (validates the `allOf`/`$ref` schema registers).

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.101",
+    "need4deed-sdk": "0.0.102",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
     "pino-pretty": "^13.0.0",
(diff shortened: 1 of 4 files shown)


need4deed (approved):
Approved: default missing language to en (SDK 0.0.102 makes ApiUserPost.language optional); schema allOf+default keeps Lang enum, handler backstop. Tests/typecheck green.

PR #649, opened 2026-06-08
fix(user): default timezone to "CET" on registration

arturasmckwcz:
## Bug

`POST /user` failed with `timezone must be a string` (class-validator `@IsString()`). The `ApiUserPost` refactor (#647) dropped `timezone` from the body/handler, but the `User` entity validates `timezone` as a required string — and `validate(newUser)` runs on the in-memory object **before** save, where the `@Column({ default: "CET" })` DB default hasn't applied yet.

## Fix

Set `timezone: "CET"` explicitly when constructing the user (server-controlled, mirrors the entity default). One line.

`yarn typecheck` + lint clean.

### DIFF
--- a/src/server/routes/user.ts
+++ b/src/server/routes/user.ts
@@ -336,6 +336,10 @@ export default async function userRoutes(
         role,
         isActive: false,
         language: language ?? Lang.EN,
+        // Server-controlled (not in ApiUserPost). Set explicitly to the entity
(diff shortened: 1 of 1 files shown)


need4deed (approved):
Approved: default timezone to 'CET' so the in-memory User passes @IsString validation. Independent of the #650 CDN hotfix; unblocks registration.

PR #660, opened 2026-06-09
fix(#659): always enforce number consistency in fuzzy agent match

nadavosa:
## Summary

- Number consistency check now runs regardless of how many fuzzy candidates exist
- Prevents "Heerstr 10" from matching form input "Heerstr 110"
- Agents with no number in title (e.g. "Refugium Hausvaterweg") still match on street name alone
- Also simplifies the multi-match logic into a single filter pass

Closes https://github.com/need4deed-org/be/issues/659

## Test plan

- [ ] Submit form with `rac_address = "Heerstr 110"` — should not match existing "Heerstr 10" agent, should create new one
- [ ] Submit form with `rac_address = "Hausvaterweg 21"` — should still match "Refugium Hausvaterweg" (no number in title)
- [ ] `yarn test -- src/test/server/utils/data/get-agent-by-postcode.test.ts`

### DIFF
--- a/src/server/utils/data/get-agent-by-postcode.ts
+++ b/src/server/utils/data/get-agent-by-postcode.ts
@@ -48,31 +48,24 @@ export function getAgentByAddress(
 
   // 2. Fuzzy fallback for legacy agents: street name (no number) found as a
   //    whole word in agent title + PLZ from either address.postcode or agentPostcode.
+  //    Number consistency: if the agent title contains a number it must match the
+  //    form's number — prevents "Heerstr 10" matching form input "Heerstr 110".
+  //    Agents with no number in title (e.g. "Refugium Hausvaterweg") match on
+  //    street name alone.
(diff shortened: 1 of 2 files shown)


arturasmckwcz (approved):
Reviewed — the fix is correct: running the house-number consistency check for every fuzzy candidate (not just when 2+) cleanly removes the "Heerstr 10" ↔ "Heerstr 110" false positive, while no-number titles still match on street name alone. The single-pass filter is a nice simplification, and the two new tests cover the rejection + the correct match.

Non-blocking note: this drops the old "narrow by number" disambiguation — a no-number agent and an exact-number agent in the same PLZ now both survive → `undefined` (was: pick the numbered one). That's consistent with the "ambiguous ⇒ don't guess" philosophy and arguably safer; a test pinning that case would be a good follow-up.

Approving.

Issue #666, opened 2026-06-10
security: search filter endpoints can infer PII-masked field values


arturasmckwcz:
## Background

Response masking was added to hide PII (Person name, Address details) from non-privileged API callers. However, the list endpoints still filter against the **unmasked database columns** in their `WHERE` clauses (`getAgentWhere`, `getVolunteerWhere`, `getOpportunityWhere`, etc.).

## Problem

A non-privileged caller can infer a masked value by probing search patterns — for example, searching for "Müller" and seeing which (masked) rows match — even though the response shows "***". This makes the masking ineffective as a PII protection measure.

Additionally, PII embedded in **free-text fields** (e.g. legacy `<|>` contact comments, communication notes) is not reached by entity-level masking at all.

## Fix

For non-privileged roles, either:
1. **Scope/disable** the searchable filters (don't allow filtering on masked fields), OR
2. **Apply the same visibility logic to the query `WHERE`** so the filter only runs if the caller would see the field

Also consider redacting known free-text PII patterns in comments and notes.

`// SECURITY:` markers have already been left at the relevant filter sites in the code.

## Acceptance criteria
- [ ] Non-privileged callers cannot infer masked field values through search
- [ ] Free-text PII in comments/notes is either masked or excluded from search for non-privileged roles
- [ ] Privileged roles (COORDINATOR, ADMIN) are unaffected

nadavosa:
Closing as part of a backlog cleanup: open 60+ days with no assignee. Reopen if this is still needed.

PR #667, opened 2026-06-10
Mask PII on GET /opportunity, /volunteer, /agent by role

arturasmckwcz:
Opens the GET endpoints under `/opportunity/*`, `/volunteer/*`, `/agent/*` to any logged-in user and **masks PII (`Person`/`Address`)** the caller isn't permitted to see. COORDINATOR/ADMIN are unaffected; writes stay COORDINATOR-only.

## Auth
- The blanket `authenticate({ role: COORDINATOR })` `onRequest` hook on the three route groups is relaxed to `authenticate()` (logged-in only).
- Every POST/PATCH/DELETE is **re-gated** per-route to `authenticate({ role: COORDINATOR })`. `/agent/membership` stays COORDINATOR; `/agent/register/search` stays public.

## Masking
- New `src/server/utils/pii/`: `maskString` (random char + `***`, e.g. `John` → `x***`), `maskPii` (entity-graph walker using `instanceof Person/Address`, cycle-safe), `resolveVisiblePersonIds`, and a `makePiiSerialization` `preSerialization` hook + `maskForCaller`/`resolveCallerMask` helpers for routes whose DTO is built in a helper.
- The hook masks the loaded **entity graph** then runs the DTO, so masked values stay strings — **no response-schema changes**.
- Caller identity reuses the user `authenticate()` already loads (exposed as `request.authUser`) — no JWT change, role stays DB-authoritative.

### Who sees an unmasked Person
| Role | Unmasked |
|---|---|
| USER | none |
| VOLUNTEER | own (`id === personId`) |
| AGENT | own ∪ their agent's members ∪ volunteers on their agent's opportunities |
| COORDINATOR/ADMIN | all |

## Validation
- 7 new unit tests for `maskString`/`maskPii`; all 291 existing tests still pass; typecheck + lint clean.
- Live smoke (role-specific access cookies): ADMIN/COORDINATOR see real names/emails; AGENT sees `Sarah`→`g***`, emails→`y***` on both the hook path (`GET /agent/:id`, `/agent`) and the inline path (`GET /volunteer/:id`, `/volunteer`); AGENT `PATCH` → 403.

## Known limitation (flagged)
Search filters run on **unmasked DB columns**, so a non-privileged caller can infer a masked value by probing matches; free-text PII (legacy `<|>` comments) isn't reached. `// SECURITY:` markers added at the filter sites; tracked in #666.

### DIFF
--- a/src/server/routes/agent/agent.routes.ts
+++ b/src/server/routes/agent/agent.routes.ts
@@ -1,11 +1,5 @@
 import { FastifyInstance, FastifyPluginOptions } from "fastify";
-import {
-  ApiAgentGet,
-  ApiAgentGetList,
-  ApiAgentPatch,
-  SortOrder,
-  UserRole,
-} from "need4deed-sdk";
+import { ApiAgentPatch, SortOrder, UserRole } from "need4deed-sdk";
 import { BadRequestError, NotFoundError } from "../../../config";
 import Agent from "../../../data/entity/opportunity/agent.entity";
 import logger from "../../../logger";
@@ -65,7 +59,9 @@ export default async function agentRoutes(
 
   fastify.get<{
     Querystring: QuerystringAgentGetList;
-    Reply: ReplyDataCount<ApiAgentGetList[]>;
+    // Handler sends entities; the DTO (ApiAgentGetList) runs in the
+    // preSerialization hook, so the send is typed as the entity.
+    Reply: ReplyDataCount<Agent[]>;
   }>(
     "/",
     {
@@ -110,17 +106,16 @@ export default async function agentRoutes(
         await agentRepository.save(updates);
       }
 
-      // DTO (dtoAgentGetList) runs in the preSerialization hook after PII
-      // masking, so the wire shape is ApiAgentGetList[] despite sending entities.
+      // DTO (dtoAgentGetList) runs in the preSerialization hook after PII masking.
       return reply.status(200).send({
         message: `Agents page:${page || 1} fetched successfully`,
-        data: agentsDistrict as unknown as ApiAgentGetList[],
+        data: agentsDistrict,
         count,
       });
     },
   );
 
-  fastify.get<{ Params: ParamsId; Reply: ReplyData<ApiAgentGet> }>(
+  fastify.get<{ Params: ParamsId; Reply: ReplyData<Agent> }>(
     "/:id",
     {
       schema: {
@@ -157,7 +152,7 @@ export default async function agentRoutes(
       // DTO (dtoAgentGet) runs in the preSerialization hook after PII masking.
       return reply.status(200).send({
         message: `Agent (id:${id}) fetched successfully`,
-        data: agentComments as unknown as ApiAgentGet,
+        data: agentComments,
       });
     },
   );
--- a/src/server/routes/agent/agent.routes.ts
+++ b/src/server/routes/agent/agent.routes.ts
@@ -36,6 +36,7 @@ import {
   patchAddress,
   updateAgentLanguages,
 } from "../../utils";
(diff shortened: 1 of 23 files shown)


src/server/routes/agent/agent.routes.ts
```diff
@@ -103,15 +105,16 @@ export default async function agentRoutes(
...
       if (updates.length > 0) {
         await agentRepository.save(updates);
       }
 
+      // DTO (dtoAgentGetList) runs in the preSerialization hook after PII
+      // masking, so the wire shape is ApiAgentGetList[] despite sending entities.
       return reply.status(200).send({
         message: `Agents page:${page || 1} fetched successfully`,
-        data,
+        data: agentsDistrict as unknown as ApiAgentGetList[],
```
arturasmckwcz: is it necessary skipping type checking here?

src/server/routes/agent/agent.routes.ts
```diff
@@ -145,22 +149,23 @@ export default async function agentRoutes(
...
 
       if (updates.length > 0) {
         await agentRepository.save(updates);
       }
 
+      // DTO (dtoAgentGet) runs in the preSerialization hook after PII masking.
       return reply.status(200).send({
         message: `Agent (id:${id}) fetched successfully`,
-        data,
+        data: agentComments as unknown as ApiAgentGet,
```
arturasmckwcz: is it necessary skipping type checking here?

src/server/routes/opportunity/opportunity-volunteer.routes.ts
```diff
@@ -44,11 +48,10 @@ export default function opportunityOpportunityVolunteerRoutes(
         ],
       });
 
-      const data = volunteers.map(opportunityOpportunityVolunteerDTO);
-
+      // DTO runs in the preSerialization hook after PII masking.
       return reply.status(200).send({
         message: `Volunteers for opportunity id:${opportunityId}.`,
-        data,
+        data: volunteers as unknown as ApiVolunteerOpportunityGet[],
```
arturasmckwcz: is it necessary skipping type checking here?

src/server/routes/agent/agent.routes.ts
arturasmckwcz: Good catch — it wasn't necessary. The cast was only there because the route still typed `Reply` as the DTO (`ApiAgentGet`…) while the handler now sends the **entity** (the DTO runs in the preSerialization hook). Fixed in 742bccf by typing `Reply` as the entity it actually sends — so the send stays fully type-checked, no `as unknown as`. The wire contract is still the response schema.

src/server/routes/agent/agent.routes.ts
arturasmckwcz: Good catch — it wasn't necessary. The cast was only there because the route still typed `Reply` as the DTO (`ApiAgentGet`…) while the handler now sends the **entity** (the DTO runs in the preSerialization hook). Fixed in 742bccf by typing `Reply` as the entity it actually sends — so the send stays fully type-checked, no `as unknown as`. The wire contract is still the response schema.

src/server/routes/opportunity/opportunity-volunteer.routes.ts
arturasmckwcz: Same as above — fixed in 742bccf: `Reply` is now typed as `OpportunityVolunteer[]` (the entity the handler sends), dropping the cast.

need4deed (approved):
Approving: review findings addressed — comment-author PII masking, document GET routes re-gated to COORDINATOR, AGENT visibility tests added. Filter side-channel tracked in #666.

PR #673, opened 2026-06-11
feat(agent): reuse getAgentByAddress picker in self-registration

arturasmckwcz:
## What

Reuses the `POST /opportunity/legacy` agent picker (`getAgentByAddress` — strict normalized street + PLZ, then fuzzy street-name-in-title with house-number consistency) across **both** agent-picking surfaces of the self-registration flow (#591/#601), replacing the naive `street ILIKE`.

### 1. `GET /agent/register/search`
- New `postcode` query param (alongside `street`).
- Loads the **domain-scoped** candidate agents (member shares the registrant's email domain — unchanged) with the relations the matcher reads, then runs `getAgentByAddress`.
- Returns the **single best match** (or none) — same picker semantics as legacy.

### 2. `POST /agent/register` (CREATE)
- `createAgentForPerson` now **dedups before creating**: if `addressStreet`+`addressPostcode` already resolve to an existing agent, it throws `AgentAddressConflictError`, which the route maps to **409 + `agentId`** (`conflict: "address"`) so the client can offer to JOIN. This is the dedup #591 deferred.

## Contract

- `registerSearchQuerySchema` gains `postcode`; `registerAgentConflictSchema.conflict` enum gains `"address"`.
- SDK: need4deed-org/sdk#126 adds `ApiAgentAddressConflict` + `ApiAgentRegisterConflict`. **The BE doesn't import the conflict type** (the 409 is validated against the local AJV schema), so no `need4deed-sdk` bump is required here — the SDK PR just keeps the published contract/Swagger in sync for the FE.

## FE coordination

The search picker now **requires** `postcode` to return matches (without it, it returns `[]`). The FE registration wizard already collects a postcode, so it should pass `&postcode=` to `/agent/register/search`.

## Notes / behaviour

- Search result shape is unchanged (array); it now contains 0 or 1 item (the best address match) rather than up to 10 ILIKE hits. `getAgentByAddress` deliberately returns nothing when a fuzzy match is ambiguous (multiple candidates) — the FE flow still allows creating.
- Create-flow dedup scans all agents (not domain-scoped) to catch any duplicate at that address; the subsequent JOIN still applies `resolveJoinStatus` (domain match → ACTIVE, else PENDING).

## Tests

`write-agent-registration.test.ts` (+1: dedup throws `AgentAddressConflictError`, no create; existing create tests default the dedup lookup to "no match"). Full suite green (317).

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.104",
+    "need4deed-sdk": "0.0.105",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
     "pino-pretty": "^13.0.0",
--- a/src/server/routes/agent/register.routes.ts
+++ b/src/server/routes/agent/register.routes.ts
@@ -22,8 +22,10 @@ import {
   responseErrors,
 } from "../../schema";
 import {
+  AgentAddressConflictError,
   classifyRegisterAgentConflict,
   createAgentForPerson,
+  getAgentByAddress,
   joinAgent,
   resolveJoinStatus,
 } from "../../utils";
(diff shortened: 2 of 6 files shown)


need4deed (approved):
Approving: registration reuses getAgentByAddress (search picker + create-flow address dedup -> 409 join); on need4deed-sdk 0.0.105; tests green.

PR #674, opened 2026-06-12
fix(migrations): make data migrations self-contained (no entity/helper/enum drift)

arturasmckwcz:
## Symptom

A from-scratch DB rebuild failed on:

```
Migration "BackfillOpportunitySubmittedByPerson4FieldBlobs1780182685021" failed,
error: column AgentPerson.status does not exist
```

## Root cause (a whole class of bug, not one migration)

Several data/seed migrations import **live entities, app helpers, config constants, and SDK enums** — all of which drift as the codebase evolves. On a fresh replay TypeORM emits SQL from the *current* shape, which may not match the schema/contract as it was at that point in history. The concrete break: the submitter backfills read the `AgentPerson` entity, which **later** gained a `status` column (a later migration), so replay selected a column that didn't exist yet.

## Fix — make every affected migration self-contained (raw SQL + hardcoded literals)

| Migration | Was | Now |
|---|---|---|
| `1780169787517`, `1780182685021` (submitter backfills) | `getOrCreateSubmitterPerson` helper + `Person`/`AgentPerson` entities | inline person find-or-create + `agent_person` link in raw SQL |
| `1766352158945` | SDK `VolunteerCommunicationType` | hardcoded values |
| `1779739039589` | SDK `OpportunityVolunteerStatusType` | hardcoded values |
| `1776699111911` | config `titleOrphanageAgent` | hardcoded |
| `1776321570996` | `logger` / `NotFoundError` imports | plain `Error`, no imports |

The backfill rewrite is **behaviour-identical**: those blobs carry no address, so the helper's address-sync was already a no-op there; and `created_at`/`updated_at`/`preferred_communication_type` have DB defaults, so the raw INSERT matches the old entity save (smoke-tested in a rolled-back txn).

Editing shipped migrations is safe — TypeORM tracks by **name**, so applied DBs don't re-run them; only fresh replays get the corrected body (the broken case).

## Docs

`CLAUDE.md`: adds the `docker compose down -v` + `RUN_MIGRATIONS=1 docker compose up` fresh-rebuild recipe as the canonical self-containment check, and states the rule (raw SQL + literals only; no entities/helpers/config/SDK enums).

## Verification

Full stack rebuilt from scratch (`docker compose down -v` → `RUN_MIGRATIONS=1 docker compose up`): all **26 pending migrations run clean** to `Migrations completed`, including the 4-field backfill processing **29 seeded rows** (12 created, 17 refreshed) via the new raw SQL — no `column ... does not exist`. typecheck ✓, lint ✓, full suite (317) ✓.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -51,6 +51,29 @@ Docker alternative (includes Postgres):
 docker compose up           # starts db + bootstrap + be
 ```
 
+### Flushing the database (fresh rebuild)
+
+To wipe the DB and replay the full migration chain from scratch — e.g. to
+verify migrations are self-contained and replay cleanly:
+
+```bash
+docker compose down -v                  # stop + drop the be_database volume
+RUN_MIGRATIONS=1 docker compose up      # db -> bootstrap (re-seeds) -> be runs all migrations
+```
+
+`down -v` removes the `be_database` volume (the data); the next `up` re-seeds
+via the `bootstrap` service, then `be` runs every migration on startup
+(`RUN_MIGRATIONS=1` forces it on regardless of `.env`). Watch the `be` logs —
+the run must reach `Migrations completed` with no `column ... does not exist`
+errors.
+
+**This is the canonical check that a data/seed migration is self-contained.**
+Migrations must depend only on **raw SQL + hardcoded literals** — never import
+live entities, app helpers, config constants, or **SDK enums** (an enum value
+can change with a contract update, retroactively altering what an old migration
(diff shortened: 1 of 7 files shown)


need4deed (approved):
Approving: data migrations made self-contained (raw SQL + hardcoded literals, no entity/helper/config/SDK-enum drift); verified by full down -v + RUN_MIGRATIONS=1 rebuild; CLAUDE.md documents the flush workflow.

PR #676, opened 2026-06-12
feat(trusted-domain): allowlist fixes brand-new agent self-registration + CRUD

arturasmckwcz:
## Bug

A brand-new agent representative couldn't self-register. The domain gate — `POST /user` (`role: agent`) and `resolveJoinStatus` — only passed when an **existing** member already shared the registrant's email domain, so the **first** rep of an org not yet in the system was rejected (`POST /user` → 404).

## Fix — a managed trusted-domain allowlist as the fallback

- **`TrustedDomain`** entity (`domain` unique) + a **self-contained** seed migration (raw SQL + hardcoded list, per the migration rule) seeding the agreed allowlist, **deduped** (`need4deed.org` and `eu-homecare.com` appeared twice → **36** unique).
- **`isEmailDomainTrusted(email)`** helper + `trustedDomainRepository` decorator.
- **Wired into both gates:**
  - `POST /user` AGENT → passes if a member shares the domain **OR** `isEmailDomainTrusted`.
  - `resolveJoinStatus` → ACTIVE on member-match **OR** trusted, else PENDING.

## Manage it — `/trusted-domain` CRUD (COORDINATOR/ADMIN)

`GET` (list) · `POST {domain}` (201, 409 on dup) · `PATCH /:id {domain}` · `DELETE /:id`. Domains normalized to lowercase; `POST`/`PATCH` validate a bare hostname pattern.

## Contract

`need4deed-sdk` `0.0.105 → 0.0.106` — `ApiTrustedDomain` / `…Post` / `…Patch` (sdk#127). `0.0.106` also carries `ApiAgentAddressConflict` (sdk#126, which was published-but-unmerged; now landed on `main`).

`POST /opportunity/legacy` is unchanged, per the agreed scope.

## Verification

- Migration runs + seeds **36** distinct domains (verified against the live DB).
- typecheck ✓, lint ✓, full suite **322** (+5: `isEmailDomainTrusted` unit tests + a `resolveJoinStatus` trusted-domain path; server-boot tests confirm the new routes/schemas register).

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.105",
+    "need4deed-sdk": "0.0.106",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
     "pino-pretty": "^13.0.0",
--- a/src/data/data-source.ts
+++ b/src/data/data-source.ts
@@ -39,6 +39,7 @@ import Skill from "./entity/profile/skill.entity";
 import Testimonial from "./entity/testimonial.entity";
 import Timeslot from "./entity/time/timeslot.entity";
 import Timeline from "./entity/timeline.entity";
+import TrustedDomain from "./entity/trusted-domain.entity";
 import User from "./entity/user.entity";
 import Appreciation from "./entity/volunteer/appreciation.entity";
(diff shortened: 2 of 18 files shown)


need4deed (approved):
Approving: trusted-domain allowlist fixes brand-new agent self-registration (POST /user AGENT gate + resolveJoinStatus fallback) + COORDINATOR/ADMIN CRUD; self-contained seed migration (36 deduped); on sdk 0.0.106; suite green (322).

PR #686, opened 2026-06-15
feat(opportunity): add volunteerNames to GET /opportunity list

arturasmckwcz:
## What

Surfaces the names of volunteers matched (m2m) to each opportunity on `GET /opportunity` (list), via `ApiOpportunityGetList.volunteerNames`.

- bump `need4deed-sdk` `0.0.106 → 0.0.107` (`volunteerNames` on `ApiOpportunityGetList`, sdk#128)
- add `volunteerNames` (optional `string[]`) to the `ApiOpportunityGetList` definition in `sdk-types.json`
- `dtoOpportunityGetList` maps `opportunityVolunteer.volunteer.person.name` (filtered to non-empty)
- load the `opportunityVolunteer.volunteer.person` relation in the `GET /opportunity` list query

## PII

Names are `person.name`, which is **masked before the DTO** by the `preSerialization` hook — so for a non-privileged caller the array contains masked names (e.g. `"x*** y***"`) except for volunteers their role may see (e.g. the owning agent). COORDINATOR/ADMIN see full names. Field is named `volunteerNames` (not `volunteers`) to avoid implying objects.

## Tests

`dtoOpportunityGetList` unit tests: maps names + skips person-less rows, masked-passthrough, empty when none. Full suite green (325, +3).

> Note: no end-to-end request test (harness has no seeded-user route rig) — covered at the DTO unit level + server-boot schema registration, as with the earlier PII PRs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.106",
+    "need4deed-sdk": "0.0.107",
     "pg": "^8.14.1",
     "pino": "^10.3.1",
     "pino-pretty": "^13.0.0",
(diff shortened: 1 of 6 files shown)


need4deed (approved):
Approving: volunteerNames on GET /opportunity list; masked per role before the DTO; verified live that the to-many relation leaves count/take correct (839 stable); on sdk 0.0.107; suite green (325).

PR #687, opened 2026-06-15
fix(opportunity): volunteerNames = only opp-matched volunteers

arturasmckwcz:
## What

Follow-up to #686: restrict `volunteerNames` on `GET /opportunity` (list) to volunteers whose `opportunity_volunteer.status` is **`opp-matched`** — excluding `opp-pending` / `opp-active` / `opp-past` links.

```ts
.filter((ov) => ov.status === OpportunityVolunteerStatusType.MATCHED)
```

BE-only — the contract (`ApiOpportunityGetList.volunteerNames: string[]`) is unchanged, so no SDK bump. The `opportunityVolunteer` relation already loads `status` (a column), so no query change either.

## Tests

`dtoOpportunityGetList` tests updated: only `opp-matched` names map (pending/active/past excluded), masked-passthrough for a matched volunteer, empty when the only link is non-matched. Suite green (325).

### DIFF
--- a/src/services/dto/dto-opportunity.ts
+++ b/src/services/dto/dto-opportunity.ts
@@ -4,6 +4,7 @@ import {
   ApiOpportunityGetList,
   ApiVolunteerOpportunityGetList,
   OpportunityType,
+  OpportunityVolunteerStatusType,
 } from "need4deed-sdk";
 import Comment from "../../data/entity/comment.entity";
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Approving: volunteerNames restricted to opp-matched volunteers; BE-only (contract unchanged); tests updated; suite green (325).

PR #713, opened 2026-06-24
✨ feat: add POST /user/admin for admin-privileged user creation

arturasmckwcz:
## Summary

- Adds `POST /user/admin` — an admin-authenticated endpoint that creates users with any role, including `admin` and `coordinator`
- Public `POST /user/` remains unchanged (still blocks privileged roles, still sends email verification)
- Admin-created accounts are set `isActive: true` immediately — no email verification flow

## Motivation

The existing `POST /user/` endpoint is for self-registration and explicitly rejects `admin`/`coordinator` roles. There was no API path to provision privileged users without direct DB access.

## Contract

Reuses existing SDK types (`ApiUserPost` body, `userResponseSchemaIncludePerson` response) — no SDK change required.

## Test plan

- [ ] `POST /user/admin` without auth → 401
- [ ] `POST /user/admin` with volunteer session → 403
- [ ] `POST /user/admin` with admin session + `role: coordinator` → 201, `isActive: true`
- [ ] `POST /user/admin` with admin session + `role: admin` → 201, `isActive: true`
- [ ] Duplicate email → 409
- [ ] `POST /user/` with `role: admin` still → 401 (unchanged)

### DIFF
--- a/src/server/routes/user.ts
+++ b/src/server/routes/user.ts
@@ -400,4 +400,89 @@ export default async function userRoutes(
       return reply.status(201).send(savedUser);
     },
   );
+
+  // Admin-only user creation — accepts any role including admin/coordinator.
+  // Unlike POST /user/ this endpoint requires an authenticated admin session
+  // and activates the account immediately (no email verification flow).
+  fastify.post<{
+    Body: ApiUserPost;
+    Reply: User | { message: string; errors?: any };
+  }>(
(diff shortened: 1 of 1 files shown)


need4deed (approved):
LGTM — clean separation from public registration, correct auth guard, isActive:true avoids orphaned unverified accounts.

PR #716, opened 2026-06-25
Change password

kiborrdis:
## Description
Implemented reset request and reset password endpoints

**Some concerns**

1) It seems like password-reset endpoint should ultimately be 2 different endpoints:
password-reset and password-change
One is for the reset password use case, and other for the change password use case. It's two different user flows, and there is no pros that I can see in combining them into one endpoint, other than "a little bit less amount of code".
If we split them:
- We will avoid confusing token vs authed user logic in the handler
- You can filter metrics/logs
- You can change and maintain the one without risk of broking the other

We can leave it as is, or if you agree to split them, say so. It's a short fix

2) There is no new password complexity validation. I looked at the register endpoint, and there are also seems to be none. Is that planned on just an overlook? It's probably better to forbid users to use simple passwords such as "11111111"

## Related Issues
Closes #582 

## Changes
- Added /auth/request-reset endpoint
- Added new file to "notify/events" for resetPassword email. It's basically the same as email verification file, please check if cdn route for reset email template is acceptable
- Added /auth/password-reset endpoint
- Added parameter "optional" in fastify.authenticate. Since we need to resolve a access cookie to a user, but if there none allow handler to run.
- Added basic tests for both endpoints

## Screenshots / Demos
-

## Checklist

[x] POST /auth/request-reset sends a reset email with a token link when the email exists
[x] POST /auth/password-reset supports both the token-based path and the password-confirmation path

Missing both token and password returns 400

### DIFF
--- a/src/server/plugins/jwt.ts
+++ b/src/server/plugins/jwt.ts
@@ -2,6 +2,11 @@ import fastifyJwt from "@fastify/jwt";
 import { FastifyInstance, FastifyReply, FastifyRequest } from "fastify";
 import fp from "fastify-plugin";
 import { UserRole } from "need4deed-sdk";
+import {
+  BadRequestError,
+  UnauthenticatedError,
+  UnauthorizedError,
+} from "../../config";
 import { accessCookieName, cookieOptions } from "../../config/constants";
 import logger from "../../logger";
 import { AuthOptions } from "../types";
@@ -34,9 +39,9 @@ async function jwtPlugin(
       try {
         try {
           await request.jwtVerify();
-        } catch (_error) {
+        } catch {
           reply.status(401);
-          throw new Error("Authorization failed.");
+          throw new UnauthenticatedError("Authorization failed.");
         }
 
         const userId = request.user?.id;
@@ -53,7 +58,7 @@ async function jwtPlugin(
 
         if (!user) {
           reply.status(404);
-          throw new Error("User not found.");
+          throw new BadRequestError("User not found.");
         }
 
         // Expose the already-loaded user (carries personId + DB-authoritative
@@ -77,14 +82,14 @@ async function jwtPlugin(
 
         if (role && role !== user.role) {
           reply.status(403);
-          throw new Error("Permission denied");
+          throw new UnauthorizedError("Permission denied");
         }
 
         if (allowSelf) {
           const requestParamId = (request.params as { id?: string }).id;
           if (String(userId) !== requestParamId) {
             reply.status(403);
-            throw new Error("Permission denied");
+            throw new UnauthorizedError("Permission denied");
           }
         }
       } catch (error) {
--- a/src/server/plugins/jwt.ts
+++ b/src/server/plugins/jwt.ts
@@ -35,9 +35,6 @@ async function jwtPlugin(
         try {
           await request.jwtVerify();
         } catch (_error) {
-          if (opt?.optional) {
-            return;
-          }
           reply.status(401);
           throw new Error("Authorization failed.");
         }
--- a/src/server/plugins/jwt.ts
+++ b/src/server/plugins/jwt.ts
@@ -35,6 +35,9 @@ async function jwtPlugin(
         try {
           await request.jwtVerify();
         } catch (_error) {
+          if (opt?.optional) {
+            return;
+          }
           reply.status(401);
           throw new Error("Authorization failed.");
         }
--- a/src/server/plugins/jwt.ts
+++ b/src/server/plugins/jwt.ts
@@ -1,13 +1,8 @@
 import fastifyJwt from "@fastify/jwt";
-import { FastifyInstance, FastifyReply, FastifyRequest } from "fastify";
+import { FastifyInstance, FastifyRequest } from "fastify";
 import fp from "fastify-plugin";
 import { UserRole } from "need4deed-sdk";
-import {
-  BadRequestError,
-  BaseError,
-  UnauthenticatedError,
-  UnauthorizedError,
-} from "../../config";
+import { UnauthenticatedError, UnauthorizedError } from "../../config";
 import { accessCookieName, cookieOptions } from "../../config/constants";
 import logger from "../../logger";
 import { AuthOptions } from "../types";
@@ -25,7 +20,7 @@ async function jwtPlugin(
   });
 
   fastify.decorate("authenticate", function (opt?: AuthOptions) {
-    return async function (request: FastifyRequest, reply: FastifyReply) {
+    return async function (request: FastifyRequest) {
       logger.debug(
         `jwtPlugin:authenticate called with request.routeOptions.config: ${JSON.stringify(request.routeOptions.config)}`,
       );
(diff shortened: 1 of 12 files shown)


src/server/types/auth.ts
```diff
@@ -3,4 +3,5 @@ import { UserRole } from "need4deed-sdk";
 export interface AuthOptions {
   role?: UserRole;
   allowSelf?: boolean;
+  optional?: boolean;
```
arturasmckwcz: why do we need that?
for open to general public endpoints we use either `{ public: true }` or do not call `onRequest` hook on  a route plugin at all.

src/server/plugins/jwt.ts
```diff
@@ -35,6 +35,9 @@ async function jwtPlugin(
         try {
           await request.jwtVerify();
         } catch (_error) {
+          if (opt?.optional) {
```
arturasmckwcz: why did you set bypassing role check if token's got problems?
pls remove that.

and if you don't mind, please refactor replying and throwing generic Error to throwing the custom `UnauthenticatedError` instead.

src/server/routes/auth.ts
```diff
@@ -236,6 +237,140 @@ async function authRoutes(
...
     },
   );
+
+  fastify.post<{
+    Body: { email: string };
+    Reply: { message: string };
+  }>(
+    prefixedPath + RoutePrefix.REQUEST_RESET,
+    {
+      schema: {
```
arturasmckwcz, line 248: please move schemas to `src/server/schema`
there are wrappers

src/server/routes/auth.ts
```diff
@@ -236,6 +237,140 @@ async function authRoutes(
...
+  );
+
+  fastify.post<{
+    Body: { token?: string; password?: string; newPassword: string };
+    Reply: { message: string };
+  }>(
+    prefixedPath + RoutePrefix.RESET_PASSWORD,
+    {
+      onRequest: [fastify.authenticate({ optional: true })],
+      schema: {
```
arturasmckwcz, line 281: `src/server/schema`

src/server/routes/auth.ts
```diff
@@ -236,6 +237,140 @@ async function authRoutes(
       return reply.status(200).send({ message: "Logout successful." });
     },
   );
+
+  fastify.post<{
+    Body: { email: string };
```
arturasmckwcz, line 243: pls check `src/server/types/endpoint-handlers.ts` for readymade generics

src/server/routes/auth.ts
```diff
@@ -236,6 +237,140 @@ async function authRoutes(
...
+
+      return reply.status(200).send({
+        message:
+          "If an account with that email exists, a reset link has been sent.",
+      });
+    },
+  );
+
+  fastify.post<{
+    Body: { token?: string; password?: string; newPassword: string };
```
arturasmckwcz: `src/server/types/endpoint-handlers.ts`

src/server/routes/auth.ts
```diff
@@ -236,6 +237,140 @@ async function authRoutes(
...
+
+        if (payload.type !== "reset") {
+          return reply.status(400).send({ message: TOKEN_ERROR_MESSAGE });
+        }
+
+        const user = await fastify.db.userRepository.findOne({
+          where: { id: payload.id },
+        });
+
+        if (!user) {
```
arturasmckwcz: 1. better throw custom `BadRequestError`
2. deal with `! user.isActive` condition as well

arturasmckwcz (commented):
1. thank you thank you thank you the code looks splendid
2. your concerns on 2 different flows in a single handler are absolutely spot on! please add a detailed comment to the issue and implement!
3. and I've noticed several flaws, it's absolutely no problem given this is the first PR
4. thank you for helping us! incredible!

src/server/types/auth.ts
kiborrdis: For change-password flow. In which we have an authed user and an old password. We need to get a user entity from an access cookie(at least to get user's hashed password) and verify that the access token is valid. 

- If we used `authenticated` as it is without the modification we would break the flow with a reset-password token, because an unauthenticated user must be able to use the endpoint in that flow
- If we use `config: { public: true }`, `authenticated` will bail before checking the access token and transforming it to the user object, and we wouldnt be able to handle "change-password" scenario
- If we didnt use `authenticated` we again wouldn't be able to handle "change-password" scenario

It was either this flag, or reimplementing an access token check and a user object fetch in the handler. 
I decided the flag because it's not that rare of a case to need an access token resolution and allowing unauthenticated users to access an endpoint at the same time. And reimplementation of an access token check didnt look advisable

But since we are splitting the endpoint, it wont be needed anymore. I'll remove the change

src/server/plugins/jwt.ts
kiborrdis: Answer is here https://github.com/need4deed-org/be/pull/716#discussion_r3480348291

src/server/routes/auth.ts
kiborrdis: I didnt quite get what is useful from that file for this case.
Replaced the response to ReplyMessage, left the rest as is. 

If you meant something else, please say

src/server/routes/auth.ts
kiborrdis: same as https://github.com/need4deed-org/be/pull/716#discussion_r3484249668

src/server/routes/auth.ts
kiborrdis: done

src/server/routes/auth.ts
kiborrdis: I am not sure what you meant about wrappers. I moved schemas to the server/schema

If you wanted something else, please send an example

src/server/plugins/jwt.ts
kiborrdis: I replaced all throws I could with UnauthenticatedError, UnauthorizedError etc. 

Note that I kept `reply.status(<status>);` as is, because current global error handler doesnt seem to send the right error statuses for this error classes(it sends 200)

src/server/plugins/jwt.ts
```diff
@@ -34,9 +39,9 @@ async function jwtPlugin(
       try {
         try {
           await request.jwtVerify();
-        } catch (_error) {
+        } catch {
           reply.status(401);
```
arturasmckwcz: the custom error will set the status

src/server/routes/auth.ts
arturasmckwcz: this is from `GET /opportunity/{id}/volunteer-linked`
```
      schema: {
        params: idParamSchema,
        response: responseSchema("ApiOpportunityVolunteerGet#", true, false),
      },
```

arturasmckwcz: if you consider it's good enough what you already did moving schema objects it's totally okay. up to your discretion

src/server/plugins/jwt.ts
kiborrdis: Yeah, it didnt work cause of try catch enveloping all this errors. I added a bypass in catch block for BlockError to not break previous behaviour.

But I would propose removing try catch block altogether, and this block inside of it(it's the only explicit one it would catch after my change)
```
const userRepository = fastify.db.userRepository;
if (!userRepository) {
  throw new Error("User repository is not initialized.");
}
```
it doesnt feel like there is a point to it anyway, since you are not doing this check in handlers for the repos, and moreover if it's not initialised, all bets are off and it's ok to return 500

src/server/routes/auth.ts
kiborrdis: Used `responseSchema("")`

src/server/plugins/jwt.ts
arturasmckwcz: and BTW please remove redundant setting status before an endpoint handler.

src/server/plugins/jwt.ts
```diff
@@ -52,8 +57,7 @@ async function jwtPlugin(
         });
 
         if (!user) {
-          reply.status(404);
-          throw new Error("User not found.");
+          throw new BadRequestError("User not found.");
```
arturasmckwcz: 404 isn't appropriate here IMHO. if already logged (baked into a jwt) user is not found then it means it was removed after login and 403 fits better

src/server/plugins/jwt.ts
arturasmckwcz: I'd be pros removing try-catch. overall the desired pattern is t route all handling of errors through fastify error handler.

src/server/plugins/jwt.ts
kiborrdis: I feel like there is no practical difference. But have changed it anyway

src/server/plugins/jwt.ts
kiborrdis: Done

src/server/plugins/jwt.ts
kiborrdis: Removed the try catch and the Error throw on `!userRepository`.

src/server/plugins/jwt.ts
arturasmckwcz: you're absolutely right, thank you!

arturasmckwcz (approved):
thank you! great job!

PR #731, opened 2026-07-02
feat: allow changing opportunity type via PATCH

nadavosa:
## Summary

- Adds `volunteerType` to `ApiVolunteerOpportunityPatch` JSON schema so Fastify accepts it in PATCH requests
- Extends `parseOpportunity` to extract `body.volunteerType` and map it to the `type` column on the `Opportunity` entity
- Coordinators can now switch an opportunity between accompanying, regular, and events types after submission

## Dependency

**Requires SDK PR https://github.com/need4deed-org/sdk/pull/138 to be merged and published first.**
Once that SDK version is on npm, upgrade `need4deed-sdk` in this repo and the TypeScript cast in `parser-opportunity-patch-data.ts` can be removed.

## Scope

Handles type changes between `regular` and `events` cleanly, and from `accompanying` to another type. Changing *to* `accompanying` when no accompanying record exists requires the coordinator to also provide `accompanyingDetails` in the same PATCH request — the existing accompanying-patch logic handles creation.

Closes https://github.com/need4deed-org/be/issues/730

## Test plan

- [ ] PATCH `/opportunity/:id` with `{ "volunteerType": "regular" }` updates the opportunity type in the DB
- [ ] PATCH with `{ "volunteerType": "events" }` also works
- [ ] PATCH without `volunteerType` field leaves the type unchanged
- [ ] Invalid values are rejected by AJV with 400
- [ ] Only coordinators can PATCH (existing auth gate)


src/services/dto/parser-opportunity-patch-data.ts
```diff
@@ -34,6 +34,7 @@ export function parseOpportunity(body: ApiOpportunityPatch) {
         numberVolunteers: body.numberVolunteers,
         info: body.description,
         infoConfidential: body.description,
+        type: body.volunteerType as unknown as OpportunityType | undefined,
```
arturasmckwcz, line 38: `.claude/shared-rules.md` says:
`API changes go through the SDK contract first`
1. you need to update `ApiOpportunityPatch` on SDK first
2. you shouldn't shortcut type checking by `as unknown as ...`

PR #742, opened 2026-07-09
feat: allow changing opportunity type

kiborrdis:
## Description
Allow changing opportunity type via PATCH /opportunity/:id

## Related Issues
Closes #730

## Changes
- Add opportunity_type to schema
- Add passthrough of opportunity_type in patch data parsing
- Add creation on new accompanying record for opportunity, if there is none and type is being changed to accompanying
- Add creation/update of accompanying record for event type, add changing of event date-time in timeslot
- Add creation of accompanying details record for event type opportunity in create opportunity requests

## Checklist
- [ ] opportunity_type changes if present in patch data
- [ ] opportunity_type remains the same if not present in patch data
- [ ] if opportunity_type is changed to accompanying, accompanying record is created
- [ ] if there is accompanying in patch data, and opportunity_type is not accompanying, record is not created 
- [ ] if changing to event type, timeslot record changes to new date/time and accompanying details created/updated with this date/time
- [ ] when creating event type opportunity new accompanying details record created with date/time.


src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -476,6 +477,17 @@ export default async function opportunityRoutes(
         `PATCH /opportunity/{id} ${JSON.stringify(parseOpportunity(request.body))}`,
       );
 
+      if (
+        request.body.opportunity_type === OpportunityType.ACCOMPANYING &&
+        !opportunity.accompanyingId
+      ) {
+        if (!request.body.accompanyingDetails) {
```
arturasmckwcz: 1. `if` in `if` and nothing else
2. if something wrong in `request.body.accompanyingDetails` it might throw 500, not 400 which would be right thing.

src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -476,6 +477,17 @@ export default async function opportunityRoutes(
...
+          throw new BadRequestError(
+            'Accompanying details are required when changing opportunity type to "accompanying".',
+          );
+        }
+      }
+
       if (opportunityObj) {
         const success = await patchEntity(
```
arturasmckwcz, line 555: this needs wrapping into a transaction

src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -530,15 +542,37 @@ export default async function opportunityRoutes(
...
+          }
+          // Only create a new accompanying record if type is being changed to accompanying and no accompanying record exists yet.
+        } else if (
+          request.body.opportunity_type === OpportunityType.ACCOMPANYING
+        ) {
+          const newAccompanying = Object.assign(
+            new Accompanying(),
+            accompanying,
+          );
+          await fastify.db.accompanyingRepository.save(newAccompanying);
```
arturasmckwcz: and this write to accompanying and the next to opportunity
should be wrapped into a transaction

arturasmckwcz (commented):
thanks a lot! good code!
I'd like to stress on the wrapping into a transaction. It most probably is derived from previous setup, but now there is real risk to have an opportunity with successfully patched type to accompanying while creating accompanying record failed and `accompanying_id IS NULL`. transaction would address that IMHO.

src/server/routes/opportunity/opportunity.routes.ts
arturasmckwcz: `if` in `if` sometimes is rather good thing for better readability, but non specific error is no good

src/server/routes/opportunity/opportunity.routes.ts
kiborrdis: Yeah, you're right, added it too absentmindedly

Added transaction, please check it uses apis we supposed to use in this project

src/server/routes/opportunity/opportunity.routes.ts
kiborrdis: I find no adequate way to add `request.body.accompanyingDetails` contents check through schemas. Added straightforward hardcoded check. 

Look at my changes. If you have a more proper solution in the context of the codebase, please share

In general it feels like this type change could be it's own separate endpoint, a thought for some potential future refactoring

arturasmckwcz: I see, no problem, this isn't critical indeed, just backlog an issue for tech debt and let it go

arturasmckwcz (approved):

yarn.lock
```diff
@@ -2956,10 +2956,10 @@ natural-compare@^1.4.0:
   resolved "https://registry.yarnpkg.com/natural-compare/-/natural-compare-1.4.0.tgz#4abebfeed7541f2c27acfb29bdbbd15c8d5ba4f7"
   integrity sha512-OWND8ei3VtNC9h7V60qff3SVobHr996CTwgxubgyQYEpg290h9J0buyECNNJexkFm5sOajh5G116RYA1c8ZMSw==
 
-need4deed-sdk@0.0.122:
-  version "0.0.122"
-  resolved "https://registry.yarnpkg.com/need4deed-sdk/-/need4deed-sdk-0.0.122.tgz#2513f94fe2beed9f0fbd7b4837c47308c13a5aac"
-  integrity sha512-GtmQBUd0jYyfUDsGhuqXVJB24ddDreHIq0guBimPoBrNb8yyHH2TM5B9ulkEQ6MGAPEuOR86PxogeuLmEBIgrw==
+need4deed-sdk@0.0.123:
+  version "0.0.123"
```
arturasmckwcz: first [sdk update](https://github.com/need4deed-org/sdk/pull/163) needs to get merged and published. it'll be 0.0.128

src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -530,16 +582,86 @@ export default async function opportunityRoutes(
...
+      ) {
+        const eventDate = getDateObj(
+          request.body.event.date,
+          request.body.event.time,
         );
-        if (!success) {
-          throw new BadRequestError(
-            "Patching accompanying failed while patching opportunity.",
+
+        if (opportunity.accompanyingId) {
```
arturasmckwcz, line 654: by flipping "accompanying" to "event" only date is patched, meaning that PII sensitive fields remains as they are.
then DTOs are `oppotunity.type` deaf and passes them to serialization and then to delivery.

not that a big deal but anyway we shouldn't serve PII with events type opportunities.

arturasmckwcz (commented):
thanks a lot
1. first need to merge and publish SDK. I'll do that tomorrow
2. the major thing is PII leaking.
3. tests are missing, but that's no big deal leave it, my claude will add them later on

src/server/routes/opportunity/opportunity.routes.ts
kiborrdis: By leaking PII you mean that accompanying details stays unchanged?

I assumed it's fine since issue has this line:
`When switching from accompanying to another type, the accompanying details row is left in place (no cascade delete needed on the BE — the FE should warn the user).`
And because before this event date/time change it too could happen when changing accompanying to `event` or `regular`

But I will fix it today

arturasmckwcz: @kiborrdis I understand the issue might be (and actually is) a bit loose in terms of PII security. thanks of fixing this is important indeed!

PR #744, opened 2026-07-10
fix: add numOpportunities to AgentList schema properties

ivannissimrch:
## Description
be#738 added `numOpportunities` to the `AgentList` schema's `required` array but not to `properties`.  That's why the count never showed up on dev. 

## Related Issues
Follow-up to be #738. Unblocks fe [712](https://github.com/need4deed-org/fe/issues/712) the agent-table opportunities column

## Changes
-  Add numOpportunities to the AgentList schema properties block 

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/server/schema/sdk-types.json
+++ b/src/server/schema/sdk-types.json
@@ -1320,6 +1320,9 @@
         "numActiveVolunteers": {
           "type": "integer"
         },
+        "numOpportunities": {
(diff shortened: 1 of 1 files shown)


arturasmckwcz (approved):
thanks for taking care!

PR #757, opened 2026-07-13
fix: pre-populate be_migrations before TypeORM reads it on fresh DB

arturasmckwcz:
## Problem

After the genesis migration landed (#756), bootstrap still failed:

```
0 migrations already loaded in the database.
74 migrations were found in the source code.
74 migrations are new migrations must be executed.
Migration GenesisBaseSchema1763036250000 has been executed successfully.
query failed: CREATE TYPE "public"."config_config_key_enum" ... type already exists
Migration "UpdateVolunteerListMv1763036250587" failed
```

**Root cause:** TypeORM builds its full migration queue (all 74 migrations) by reading `be_migrations` *once* before running anything. Genesis then creates all enums/tables and inserts 73 records, but TypeORM already has a fixed plan to run every subsequent migration — they conflict with schema genesis just created.

## Fix

Add `bootstrapFreshDb()` in `src/data/index.ts`. It runs **before** `dataSource.runMigrations()`:

1. Creates `be_migrations` table if absent.
2. Counts rows — if `cnt > 0` (existing DB) returns immediately (no-op).
3. On a fresh DB: runs `GenesisBaseSchema1763036250000.up()` directly (creates full schema + inserts 73 migration records), then inserts genesis's own record.

When TypeORM then calls `runMigrations()` it reads 74/74 migrations as already done and runs zero.

## Test plan

- [ ] All 422 tests pass locally ✓
- [ ] Deploy to AITS: `kubectl rollout restart deployment/be -n n4d-dev` → bootstrap pod completes without error
- [ ] Bootstrap logs show "Genesis bootstrap completed" and no migration failures
- [ ] Existing production DB unaffected (bootstrapFreshDb exits immediately when cnt > 0)

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -1,6 +1,7 @@
 import { isProd, shouldRunMigrations } from "../config";
 import logger from "../logger";
 import { dataSource } from "./data-source";
+import { GenesisBaseSchema1763036250000 } from "./migrations/1763036250000-genesis-base-schema";
 
 const lockNumber = 0x639b4e2a1c8d79a9n; // random BIGINT (for PostgreSQL)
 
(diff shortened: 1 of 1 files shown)


need4deed (approved):
Approved.

PR #765, opened 2026-07-14
fix: notification emails silently failing across opportunity matching/creation

arturasmckwcz:
## Summary

- "Suggest" email never sent on `OpportunityVolunteer` creation — the FE creates it via a single `POST` already at `status: opp-pending`, but the send side-effect only lived in the `PATCH` transition handler. Extracted into `triggerEmailSuggestion()`, now called from both `POST` and `PATCH`.
- "Matched"/accompanying-matched emails (`email-introduction.ts`, `email-accompany-match.ts`) threw on `opportunity.contactPerson!.name`/`.phone` whenever `contactPerson` was null but `submittedByPerson` was set — a valid, common state post-`submitted_by_person` backfill. Failures were swallowed into `logger.error` after the HTTP response already returned, so they were invisible.
- New-opportunity confirmation emails (`email-new-regular.ts`, `email-new-accompanying.ts`) **always** threw, since the modern (non-legacy) opportunity-creation route never sets `contactPersonId` at all — only `submittedByPersonId`.
- Added `getOpportunityRepresentativePerson()` (`submittedByPerson → contactPerson → agent.representative.person`) and wired it into all four notify events, replacing inconsistent ad-hoc fallbacks. Updated the relations loaded before each send so the fallback chain actually has data.
- Added a pre-check + `409 ConflictError` for duplicate `(opportunityId, volunteerId)` suggestions on `POST` (previously an unhandled `500` from the DB unique constraint).
- Removed a redundant `if (id <= 0)` param check in the m2m route — already enforced by `idParamSchema`'s `minimum: 1`.

Fixes #764.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean on touched files
- [x] `yarn test:run` — 427/427 passing
- [x] Manually verified end-to-end against local `docker compose` stack: created an OV via POST (suggest) → `first-inquiry-sent` communication logged; flipped to `opp-matched` → `matched` communication logged
- [x] Verified duplicate-suggest now returns `409` instead of an unhandled `500`

### DIFF
--- a/src/data/utils/get-opportunity-representative-person.ts
+++ b/src/data/utils/get-opportunity-representative-person.ts
@@ -1,17 +1,24 @@
 import Opportunity from "../entity/opportunity/opportunity.entity";
 import Person from "../entity/person.entity";
 
-// Resolves the single person to notify/contact for an opportunity: the
-// submitter, else the legacy contact person (kept for opportunities that
-// predate `submittedByPerson`), else the opportunity's agent's current
+// Resolves the person to notify/contact for an opportunity: the submitter,
+// else the legacy contact person (kept for opportunities that predate
+// `submittedByPerson`), else the opportunity's agent's current
 // representative. Requires `submittedByPerson`, `contactPerson`, and
 // `agent.agentPerson.person` to be eager-loaded as needed by the caller.
+//
+// Prefers whichever candidate actually has an email over just the first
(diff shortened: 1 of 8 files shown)


need4deed (approved):
Verified: postcode/notification fixes tested end-to-end against local docker compose stack (opportunity creation for agent_id=1, suggest/match flow, communication log entries confirmed).

PR #770, opened 2026-07-15
fix: persist a contact person's agent role via PATCH /person/:id

arturasmckwcz:
## Summary

Fixes #767, unblocked by need4deed-org/sdk#158.

`PATCH /person/:id` silently dropped `role` because `dtoParsePerson` only ever handled `Person` columns, and role lives on `AgentPerson` (`agent_person`), not `person`. This wires up the newly-published `ApiRepresentativePatch` contract (`need4deed-sdk@0.0.126`) to actually persist it.

- `Body` type changed from `ApiPersonPatch` to `ApiRepresentativePatch` (`ApiPersonPatch & { role?, agentId? }`). When `role` is present, the handler resolves the matching `AgentPerson` row via `personId` + `agentId` and updates its role.
- `agentId` is required alongside `role` — a person can belong to more than one agent, so the target row can't be inferred from `personId` alone (this was flagged in #767's own analysis).
- `getAgentPersonRepresentative` gained an optional `agentId` parameter (backward compatible — existing personId-only callers in `user.ts`/`post.routes.ts` are unaffected) and is reused here rather than writing a new lookup.
- No changes to the GET/serialize side: `Agent.representative` (the getter already used by `dtoSerializeAgent`) reads from the same `AgentPerson` row this PATCH updates, so the new role value round-trips correctly on the next agent fetch without any DTO changes.

### Also fixed, found while implementing

- **`PATCH /person/:id` had no authentication guard at all** — not `onRequest`, not even an in-handler check. Any request, authenticated or not, could update any person's profile by ID. Added the same self-or-ADMIN authorization `GET /:id` already uses (via the modern `request.authUser` idiom rather than `GET /:id`'s legacy re-query), since this PR adds a more sensitive capability (agent role mutation) on top of this handler and shipping that onto a wide-open endpoint would have made the existing hole worse.
- **A role-only patch body crashed.** `updatePerson` unconditionally calls `personRepository.save()`; when a request contains only `role`/`agentId` (a legitimate case — "just update my role, don't touch my profile"), every `Person` column ends up unchanged and TypeORM throws `UpdateValuesMissingError` on an empty `UPDATE` SET clause. Now the person-field update is skipped entirely when the body carries no actual person fields.

## Test plan

- [x] New test file `src/test/server/routes/person.routes.test.ts` (no prior route-level test existed for this file) — covers: unauthenticated → 401, non-self/non-admin → 403, role without agentId → 400, unknown agentId → 404, self-edit persists role + survives reload (re-read from DB), admin can edit another person's role.
- [x] `yarn typecheck` passes
- [x] `yarn lint` clean on all changed/added files
- [x] Full suite: 433/433 passing (`docker compose exec be yarn test:run`)

## Not in scope for this PR

- Whether `role` should also be directly exposed on `ApiPersonGet`/`dtoSerializePerson` — not needed, since the existing agent-GET path already reads the updated value from the same row.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.125",
+    "need4deed-sdk": "0.0.126",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/src/server/routes/person.routes.ts
+++ b/src/server/routes/person.routes.ts
@@ -1,7 +1,11 @@
 import { validate } from "class-validator";
 import { FastifyInstance, FastifyPluginOptions } from "fastify";
-import { ApiPersonGet, ApiPersonPatch, UserRole } from "need4deed-sdk";
-import { NotFoundError } from "../../config";
+import { ApiPersonGet, ApiRepresentativePatch, UserRole } from "need4deed-sdk";
+import {
+  BadRequestError,
+  NotFoundError,
+  UnauthorizedError,
+} from "../../config";
 import Person, { PersonCreateType } from "../../data/entity/person.entity";
 import logger from "../../logger";
 import { dtoParsePerson, dtoSerializePerson } from "../../services";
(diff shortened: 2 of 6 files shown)


need4deed (approved):
Reviewed independently (see review comment): auth-guard fix and updatePerson empty-diff fix verified by reproduction. One accepted limitation flagged — multi-role-per-agent ambiguity mirrors existing read-side precedence (Agent.representative), not a regression. Approving.

PR #773, opened 2026-07-16
fix: unmask an agent's own address for that agent's own caller

arturasmckwcz:
## Summary

- The `maskPii` walker's `Person` branch checks visibility before masking, but the standalone `Address` branch masked unconditionally with no visibility check — and there was no `Agent` branch, so `agent.address` was always reached and masked as a plain standalone `Address`, even when the caller was viewing their own agent profile.
- Added an `Agent` branch to the walker, mirroring the `Person` branch: if `ctx.agentIds.has(node.id)`, the agent's own address is claimed via the `seen` set so the standalone-`Address` fallback doesn't re-mask it. Agents the caller has no visibility into still fall through to the standalone-`Address` branch and get masked as before.

Fixes #752

## Test plan

- [x] `yarn typecheck` — clean
- [x] `yarn eslint src/server/utils/pii/mask.ts src/test/server/utils/pii/mask.test.ts` — clean
- [x] `yarn test:run -- src/test/server/utils/pii/mask.test.ts` — 18/18 passing (16 existing + 2 new: agent's own address unmasked when `agentIds` contains it, masked otherwise)
- [x] Full suite: 424 passed; the only 3 failures are pre-existing DB-connectivity failures (no reachable local Postgres in this environment), unrelated to this change

### DIFF
--- a/src/server/utils/pii/mask.ts
+++ b/src/server/utils/pii/mask.ts
@@ -2,6 +2,7 @@ import { EntityTableName, UserRole } from "need4deed-sdk";
 import Comment from "../../../data/entity/comment.entity";
 import Address from "../../../data/entity/location/address.entity";
 import Accompanying from "../../../data/entity/opportunity/accompanying.entity";
+import Agent from "../../../data/entity/opportunity/agent.entity";
 import Opportunity from "../../../data/entity/opportunity/opportunity.entity";
 import Person from "../../../data/entity/person.entity";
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Reviewed — Agent branch correctly mirrors the Person branch's visibility check, traced end-to-end through resolveCallerVisibility/agentIds, typecheck/lint clean vs. develop baseline, 18/18 mask.test.ts passing. Approving.

PR #775, opened 2026-07-17
fix: resolve dashboard opportunity-create activities/skills/languages by id

arturasmckwcz:
## Summary
- `POST /opportunity/` (dashboard create form) reused `dealParserOpportunity`, the parser built for `POST /opportunity/legacy`, which resolves activities/skills by title text and languages by ISO code.
- The dashboard's typed create form actually sends numeric option ids (from `GET /option/activity|skill|language`), so every lookup silently returned nothing — created opportunities ended up with **zero** saved activities, skills, and languages (including the accompanying refugee language). Fixes #774.
- `need4deed-sdk@0.0.127` gives `POST /opportunity/` its own id-based contract (`languageIds`/`activityIds`/`skillIds`/`districtIds: number[]`), replacing the `string[]` fields it previously inherited from the legacy form's type.
- Added `dealParserOpportunityCreate`, which resolves those ids directly via `findBy({ id: In(ids) })`. `dealParserOpportunity` / `getProfileEntityByTitle` / `getLanguageTitle` and `POST /opportunity/legacy` are untouched.
- Extracted the timeslot-building logic (shared by both forms, unaffected by this bug) into `buildDealTimeslots` so it isn't duplicated between the two parsers.

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint` (no new errors/warnings vs. base)
- [x] `yarn test:run` — 436/436 passing, including new `opportunity-create.routes.test.ts` covering activities/skills/languages resolved by id end-to-end through the real route + DB
- [x] Manually confirmed `/opportunity/legacy` code paths (title/ISO-code parsers) are untouched

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.126",
+    "need4deed-sdk": "0.0.127",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -32,7 +32,7 @@ import {
   parseOpportunity,
   parseOpportunityLegacy,
 } from "../../../services";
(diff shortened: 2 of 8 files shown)


need4deed (approved):
Reviewed — fix and approach look correct, verified typecheck/lint clean against the same baseline as develop. Approving.

PR #778, opened 2026-07-20
Validate main-communication language is German, German+English, or none

arturasmckwcz:
## Summary

- The `PATCH /opportunity/:id` endpoint accepted any language combination for `languagesMain` (the GENERAL-purpose "main communication" language on a deal), with no server-side check.
- Adds `assertValidMainCommunicationLanguages` in `parser-opportunity-patch-data.ts`, called from the PATCH handler before any writes, resolving the submitted language ids to isoCodes via `languageRepository` and rejecting anything other than: none, German alone, or German+English (throws `BadRequestError`).
- Note: the separate `POST /opportunity` create endpoint doesn't currently preserve a main/residents distinction at all (everything in `languageIds` is flattened and stored as `LangPurpose.RECIPIENT`), so this validation only applies to the edit/patch path where the GENERAL-purpose distinction actually exists.

Pairs with need4deed-org/fe#829, which restricts the same field's dropdown to German/English and adds equivalent client-side validation.

## Test plan

- [x] `yarn typecheck` passes
- [x] `yarn lint` passes on changed files
- [ ] Manual/API check: `PATCH /opportunity/:id` with `languagesMain` containing only an English language id now returns 400; German-only and German+English still succeed

### DIFF
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -33,6 +33,7 @@ import {
   parseOpportunityLegacy,
 } from "../../../services";
 import { dealParserOpportunityCreate } from "../../../services/dto/parser-deal-opportunity-create";
+import { assertValidMainCommunicationLanguages } from "../../../services/dto/parser-opportunity-patch-data";
 import {
   idParamSchema,
   opportunityCreateBodySchema,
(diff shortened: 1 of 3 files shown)


need4deed (approved):
Reviewed — unresolved-id gap fix is correct, test coverage is solid, and I resolved a merge conflict against develop (import-order clash with #781) before verifying mergeable/typecheck/lint/full-suite (450/450) green. Safe to merge.

PR #781, opened 2026-07-20
feat: allow AGENT to update the status of their own opportunity

arturasmckwcz:
## Summary
- `PATCH /opportunity/:id` was COORDINATOR-only; an agent had no way to mark their own opportunity inactive once it's no longer relevant.
- AGENT is now allowed too, gated by `agentPersonRepository` membership on the opportunity's `agentId` (the same ownership check already used on `POST /opportunity/`).
- Restricted to the `statusOpportunity` field only — the full edit surface (contact/agent reassignment, description, etc.) stays COORDINATOR/ADMIN-only, per the security consideration raised in the issue.

Companion FE change (own PR): shows the "Change status" button for an agent viewing their own opportunity.

## Test plan
- [x] `yarn typecheck` passes
- [x] New integration test `src/test/server/routes/opportunity.routes.test.ts` covers: agent updates own opportunity's status (204), agent blocked on another agent's opportunity (403), agent blocked from patching other fields (403), coordinator full-surface patch still works (204)
- [x] Full suite (`yarn test:run`): 439/439 passing, no regressions

Fixes #820

### DIFF
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -115,7 +115,8 @@ export default async function opportunityRoutes(
   _options: FastifyPluginOptions,
 ) {
   // GETs open to any logged-in user (PII masked per role); writes stay
-  // COORDINATOR-only (re-gated per-route).
+  // COORDINATOR-only (re-gated per-route), except PATCH /:id which also lets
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Reviewed — authorization scoping is sound (field-restriction check is safe against AJV defaults/additionalProperties, per verification). No blocking issues.

PR #782, opened 2026-07-21
feat: expose all agent contacts + add POST /agent/:id/contact

arturasmckwcz:
## Summary
Part of the multi-repo effort for need4deed-org/fe#810 (support multiple contact persons on NGO/agent profile). Second step, following need4deed-sdk 0.0.130 (adds `contacts: ApiAgentMembership[]` + `ApiAgentContactPost`) — `fe` consumption is the next repo boundary.

- `dtoAgentGet` now serializes `contacts`: every `AgentPerson` membership for the agent (representative included), reusing the existing `dtoSerializeAgentMembership` helper already used for the membership-moderation endpoints (`GET /agent/membership`). `representative` is untouched — additive, not a breaking change.
- New `POST /agent/:id/contact` creates a brand-new contact (`Person` + `AgentPerson`) on an existing agent — distinct from `POST /agent/register`, which only ever links the *authenticated caller's own* person. This is a genuinely new capability: a contact created this way doesn't need a login of its own.
- Authorization: any **ACTIVE** member of that agent (mirrors the ownership check already used on `PATCH /opportunity/:id` for AGENT-role status changes — same `agentPersonRepository.findOneBy({ agentId, personId })` pattern), or COORDINATOR/ADMIN for any agent.
- Person + AgentPerson are written in one transaction (mirrors `createAgentForPerson` in `write-agent-registration.ts`).
- Address is best-effort: if street+postcode don't resolve to a known Postcode, the contact is still created without an address.
- Bumped `need4deed-sdk` to `0.0.130`.

## Not in this PR
- `fe` UI (next repo boundary — will follow once this is reviewed).
- Edit/remove for existing contacts (issue's own suggestion: "at minimum, adding should work").

## Test plan
- [x] `yarn typecheck`
- [x] `yarn test -- src/test/services/dto/dto-agent.test.ts` — added cases for the new `contacts` field (full-membership-list mapping + empty-array default), all 14 tests pass
- [x] `yarn lint` — no new errors (confirmed the 22 pre-existing errors are unrelated to any file touched here — migrations/eslint.config.js/etc.)
- [ ] No live-DB route/integration test run for the new endpoint — this sandbox's Postgres route-level integration tests (`opportunity.routes.test.ts`, `person.routes.test.ts`, `opportunity-create.routes.test.ts`) hang on `beforeAll`/`createServer()` even on an unmodified `develop` checkout (verified via a throwaway worktree at the same base commit), so this is a pre-existing environment limitation here, not something this change caused — but it also means `POST /agent/:id/contact` itself hasn't been exercised against a real DB in this sandbox. Please run the suite in CI/a working local setup before merging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.128",
+    "need4deed-sdk": "0.0.130",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/src/server/routes/agent/agent.routes.ts
+++ b/src/server/routes/agent/agent.routes.ts
@@ -34,6 +34,7 @@ import { makePiiSerialization } from "../../utils/pii/pre-serialization";
 import agentCommunicationRoutes from "./agent-communication.routes";
 import agentOpportunityRoutes from "./agent-opportunity.routes";
 import agentVolunteerRoutes from "./agent-volunteer.routes";
+import agentContactRoutes from "./contact.routes";
 import agentMembershipRoutes from "./membership.routes";
 import agentRegisterRoutes from "./register.routes";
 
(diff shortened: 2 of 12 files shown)


need4deed (approved):
Approving after live verification against the real dev DB (aits.need4deed.org) via a local tunnel: confirmed contacts list on GET /agent/:id, POST /agent/:id/contact success for an active member, 403 for a non-member, coordinator bypass working, and the role-allowlist fix (1265a040) is in place. Test rows created during verification were cleaned up afterward.

PR #783, opened 2026-07-21
fix: populate isoCode for LANGUAGE options

arturasmckwcz:
## Summary
Fixes need4deed-org/fe#839 (Main communication language dropdown shows zero options — regression from #829).

**Stacked on #782** — this branch merges in `810-agent-multiple-contacts` (#782) because bumping `need4deed-sdk` past 0.0.130 requires the `contacts` field #782 already added to `dtoAgentGet`; SDK versions are cumulative and 0.0.131 (needed here) is 0.0.130 + this fix's `isoCode` addition. GitHub will show #782's diff here until it merges — please review/merge #782 first, or review this diff with that in mind.

**Root cause**: fe's `getMainCommunicationLanguageOptions` already prefers matching `isoCode` over `title` (added as review feedback on #829's original PR) — but `GET /option/language` never actually included `isoCode`, only the translated `title`. Since fe calls that endpoint with no `language` query param, `be` defaults it to German, so `title` is always something like "Deutsch"/"Englisch" — never the literal English word "german"/"english" the fallback match was checking for. Result: the isoCode branch never fires (always undefined) and the fallback never matches (title is never literally English words) → zero options, always.

**Fix**: `getOptions`'s `getList` now looks up each `LANGUAGE` option's stable `isoCode` from the `Language` entity (via the same `entityId` already used for its translation) and includes it in the response. Also added `isoCode` to the `OptionItem` AJV schema (`additionalProperties: false` would otherwise silently strip it). Bumped `need4deed-sdk` to `0.0.131` (need4deed-org/sdk#168 — adds the `OptionItem.isoCode` field).

## Not in this PR
- No fe changes needed — the isoCode-preferring code already exists in `getMainCommunicationLanguageOptions`; it just needs real data, which this PR supplies.

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint` — no new issues (22 pre-existing errors, none in files touched here)
- [ ] No new test added for `getOptions` (no existing test file for it) and no live-DB verification — same pre-existing sandbox limitation as #782 (Postgres-backed route/integration tests hang on `beforeAll` even on unmodified `develop`, confirmed via a throwaway worktree). Please verify `GET /option/language` now returns `isoCode: "de"/"en"` for German/English in CI or a working local setup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.128",
+    "need4deed-sdk": "0.0.131",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.128",
+    "need4deed-sdk": "0.0.130",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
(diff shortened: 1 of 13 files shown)


need4deed (approved):
Approving after live verification: GET /option/language on the real dev DB now returns isoCode: "de"/"en" for German/English, confirming the fix actually reaches the response through the option-lists.json -> OptionItem# schema chain.

PR #784, opened 2026-07-21
feat: add PATCH /agent/:id/contact/:membershipId, extend create fields

arturasmckwcz:
## Summary
Part of the multi-repo effort for need4deed-org/fe#848 (unify NGO contact rows: every contact — primary and additional — gets identical fields and is individually editable, not just the primary).

- `POST /agent/:id/contact` now also accepts `middleName` and `landline`, matching the primary contact's existing field set (previously "add" and "edit" had different shapes: add had address but no middleName/landline, edit had the reverse).
- New `PATCH /agent/:id/contact/:membershipId` updates an existing contact's Person fields and/or role directly by membership — needed because `PATCH /person/:id` is strictly self-only (`authUser.personId !== personId` → 403 for anyone but ADMIN), so there was no way for an agent to edit a colleague's contact info at all. Authorized the same way `POST` already is: any **ACTIVE** member of that agent, or COORDINATOR/ADMIN for any agent.
- Partial update semantics: only fields present in the body are touched. Address is create-or-patch (an existing address is patched in place, a person with none gets a new one), mirroring the pattern `PATCH /agent/:id` already uses for the agent's own address.
- Extracted the shared role-allowlist + active-membership check (previously duplicated only in the POST handler) into `assertCanManageContacts`, used by both routes.
- Bumped `need4deed-sdk` to `0.0.132` (need4deed-org/sdk#170 — extends `ApiAgentContactPost`, adds `ApiAgentContactPatch`).

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint` — no new issues (22 pre-existing errors, none in files touched here)
- [x] **Live-verified end-to-end** against the real dev DB (`aits.need4deed.org`, via SSH tunnel), as a real agent user:
  - POST with `middleName`/`landline` → 201, both persisted correctly.
  - PATCH with only `{phone}` → only phone changed, every other field (name, email, landline, role) untouched.
  - PATCH with `{role, addressStreet, addressPostcode}` → role updated, new address created (person had none).
  - PATCH with `{addressStreet}` again → same address row patched in place (same id, no duplicate row created).
  - Non-member of the agent → 403 on PATCH.
  - Coordinator, correct agent id → 200 (bypass works).
  - Coordinator, membership under a *different* agent id than the URL → 404 (not silently accepted).
  - `GET /agent/:id` afterward showed the fully accumulated updates correctly in the contacts list.
  - Test data cleaned up afterward; confirmed the dev DB returned to its exact prior state.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.131",
+    "need4deed-sdk": "0.0.132",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/src/server/routes/agent/contact.routes.ts
+++ b/src/server/routes/agent/contact.routes.ts
@@ -19,13 +19,10 @@ import { createAgentContact, updateAgentContact } from "../../utils";
 
 // Mirrors the role allowlist on PATCH /opportunity/:id: only these three
 // roles may reach the membership check, regardless of whether a stray
-// AgentPerson row exists for some other role. Coordinator/admin may act on
-// any agent; an AGENT must be an active member of this specific agent.
-async function assertCanManageContacts(
-  fastify: FastifyInstance,
-  request: FastifyRequest,
-  agentId: number,
-): Promise<void> {
+// AgentPerson row exists for some other role. Cheap (no DB), so it runs
+// before the agent-existence check — an unauthenticated/wrong-role caller
+// shouldn't get a DB round trip just to be told "no".
+function assertHasContactManagementRole(request: FastifyRequest): void {
   const role = request.authUser?.role;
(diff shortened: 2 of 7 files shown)


need4deed (approved):
Reviewed. Solid implementation: extracted auth check avoids duplication, partial-update semantics on Person/address are correct (uses \`.update()\` with partial objects, not an ambiguous \`.save()\`), address create-or-patch mirrors the existing \`PATCH /agent/:id\` pattern, and the response re-fetches the person with relations so the address comes back populated.

One real issue from the first pass (role/membership check running before the agent-existence 404, silently changing POST's error code for a non-member-hits-nonexistent-agent edge case) has been fixed and live-verified: 404 is restored for that case, existing 403 paths are unaffected, happy path still works.

Live-verified end-to-end against the real dev DB (POST field extensions, PATCH partial update, PATCH address create-or-patch, role update, auth edge cases including the ordering fix) — test data cleaned up afterward.

Approving.

PR #786, opened 2026-07-23
fix: main communication language must be German, English, or both

nadavosa:
## Summary
Relaxes `assertValidMainCommunicationLanguages` per product decision: German alone, English alone, or both are all valid now — there's no requirement that German be included. Only a language outside the German/English set (or an unresolvable id) is rejected.

## Context
This should have shipped alongside `be#778`/`be#783`, but the paired frontend fix (`fe#849`) got closed as a duplicate of `fe#847` (which only fixed the "always rejects a valid pick" bug, not the business-rule relaxation), and this backend commit was never pushed at the time. Redone now, rebased on current `develop` (includes `#784`/`#785`).

## Test plan
- [x] `yarn typecheck`
- [x] Targeted test file: `yarn test -- src/test/services/dto/parser-opportunity-patch-data.test.ts` — 20/20 passing (updated "rejects English alone" → "resolves when the selection is English only")
- [ ] Full suite not run in this environment (no local Postgres/Docker available) — the 20 targeted tests cover the changed function directly


need4deed (approved):
Reviewed — see PR discussion. LGTM.

PR #791, opened 2026-07-24
fix: make yarn migration:run self-heal on a fresh database

arturasmckwcz:
## Summary
- `yarn migration:run` used the raw TypeORM CLI, which computes the pending-migrations list before genesis's self-registration commits — on a genuinely fresh database it still tries to replay migrations genesis already superseded and fails on the first one after it (`AddConfigDropVolunteerListMv1763036250587`, `type "config_config_key_enum" already exists`).
- `bootstrapFreshDb()` already solves this for the real app startup path (`initDatabase()`), but that logic wasn't reachable from the CLI.
- Exported `bootstrapFreshDb` and added `src/data/run-migrations.ts`, which runs it before `dataSource.runMigrations()`. Repointed the `migration:run` script at this wrapper instead of the bare `typeorm migration:run`.

## Test plan
- [x] Full fresh rebuild (`docker compose down -v && RUN_MIGRATIONS=1 docker compose up`) — genesis self-heals correctly, migrations apply cleanly, seed data populates fully
- [x] `yarn migration:run` verified idempotent on an already-migrated database ("No migrations are pending")
- [x] Full test suite passes: 454/454 (`yarn test:run`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "typeorm": "typeorm-ts-node-commonjs -d src/data/data-source.ts",
     "migration:generate": "yarn typeorm migration:generate",
     "migration:create": "yarn typeorm-ts-node-commonjs migration:create",
-    "migration:run": "yarn typeorm migration:run",
+    "migration:run": "ts-node src/data/run-migrations.ts",
     "migration:revert": "yarn typeorm migration:revert",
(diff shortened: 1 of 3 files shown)


need4deed (approved):
Verified: fresh-DB replay + full test suite (454/454) pass with this fix.

PR #796, opened 2026-07-24
feat: normalize Agent.type/services into translated reference data

arturasmckwcz:
## Summary
Part of #794. Depends on `need4deed-sdk@0.0.136` (need4deed-org/sdk#177).

- New `AgentType`/`Service` reference entities (mirroring `Skill`'s pattern: id + unique title + non-persisted `translation` prop) and an `AgentService` m2m join entity (mirroring `DealSkill`).
- `Agent.type` (enum column) → `agent.agentTypeId` FK; `Agent.services` (`text[]`) → `agent_service` join rows.
- Migration (`1784898817413-add-agent-type-and-service`):
  - Creates `agent_type`/`service`/`agent_service` tables.
  - Seeds `agent_type`/`service` from literal values snapshotted from the SDK's `AgentTypeKey`/`AgentServiceType` enums — never imported, per the self-contained-migration rule in CLAUDE.md.
  - Backfills `agent_type_id` and `agent_service` from the old columns before dropping them.
  - Seeds `field_translation` (en/de) and registers the new rows with the generic `GET /option` list registry (`option` table), then drops the old `type`/`services` columns.
- Since migrations run *before* the dev/test seed pipeline, `language` rows don't exist yet on a from-scratch bootstrap — the migration's own `field_translation` insert seeds 0 rows there. `seedFieldTranslation` gained `seedAgentTypes`/`seedServices` (mirroring `seedSkills`) as the fallback for that case. On any already-migrated environment (staging/prod, or any dev DB that predates this PR), the migration's insert does the job directly since `language` already exists there.
- `dto-agent.ts`/`parser-agent.ts`/`get-agent-where.ts`/`write-agent-registration.ts` updated for the new relations. Write payloads use `typeId`/`serviceIds` (bare ids), matching the existing `districtId` convention — confirmed with the repo owner before implementing (see sdk#177's resolved open questions).
- `GET /option` now serves `agent_type`/`service` alongside the existing skill/activity/district/etc. lists.
- Removed dead `AgentOperatorType`/`AgentType` enums in `data/types/enums.ts` (unused, and the latter's values didn't even match the real SDK enum).

**Note on translation copy:** the en/de labels seeded here are pulled from what's already shipped in `fe` (`AGENT_TYPE_LABELS`/`AGENT_SERVICE_LABELS` + the `filters.type`/`filters.services` i18n keys), not invented. Flagging in case there's a preferred/updated copy to use instead before merge.

## Test plan
- [x] `yarn typecheck` / `yarn lint` pass
- [x] Full test suite: 54 files / 464 tests pass
- [x] Canonical fresh-replay check (`docker compose down -v && RUN_MIGRATIONS=1 docker compose up`): migration applies cleanly on a genuinely empty DB, seed pipeline backfills translations, no errors
- [x] Verified `GET /option/agent_type` and `GET /option/service` return correctly translated data end-to-end (en and de)
- [ ] `be`/`fe` DTO-level real per-request translation resolution (currently the DTO passes through the loaded relation's raw title as a placeholder, matching the existing `district` shortcut in the same file) — flagged as a known follow-up in the original issue, not part of this slice

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.133",
+    "need4deed-sdk": "0.0.136",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/src/data/data-source.ts
+++ b/src/data/data-source.ts
@@ -18,6 +18,7 @@ import ActivityLog from "./entity/m2m/activity-log.entity";
 import AgentLanguage from "./entity/m2m/agent-language";
 import AgentPerson from "./entity/m2m/agent-person";
 import AgentPostcode from "./entity/m2m/agent-postcode";
+import AgentService from "./entity/m2m/agent-service";
 import CommentPerson from "./entity/m2m/comment-person";
 import DealActivity from "./entity/m2m/deal-activity";
 import DealDistrict from "./entity/m2m/deal-district";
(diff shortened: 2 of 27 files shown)


need4deed (approved):
Reviewed (including a follow-up correctness pass that found and fixed a services-filter duplicate/inflated-count bug and a translation-resolution gap, both verified live against the running dev DB). Migration replay, backfill against real pre-existing data, and the full test suite all verified. Approving.

PR #798, opened 2026-07-25
Match existing agents by street prefix in registration picker

arturasmckwcz:
GET /agent/register/search only offered a match once the full street value matched exactly (or, for legacy agents with no address row, via a fuzzy title match). Add getAgentsByStreetPrefix so partial input (3+ chars) surfaces agents with a real address.street as candidates, unioned with the existing exact/legacy match into the response list.

Closes need4deed-org/be#797

## Description

## Related Issues

## Changes

## Screenshots / Demos

## Checklist
- [ ] WITHIN THE SCOPE OF AN ISSUE; No unnecessary files included
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/server/routes/agent/register.routes.ts
+++ b/src/server/routes/agent/register.routes.ts
@@ -26,6 +26,7 @@ import {
   classifyRegisterAgentConflict,
   createAgentForPerson,
   getAgentByAddress,
+  getAgentsByStreetPrefix,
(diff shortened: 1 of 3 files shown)


need4deed (approved):
Ordering fix and added tests confirmed. LGTM.

PR #802, opened 2026-07-27
Redesign agent-search picker: drop postcode, address-prefix/title-contains matching

arturasmckwcz:
## Description

Redoes the `GET /agent/register/search` picker from #797/#798. That version required a postcode before it would search, only ever surfaced a single "decisive" best-guess match, and its legacy fallback used word-boundary regex + house-number disambiguation. Manual testing surfaced real gaps (missed a hyphenated legacy title — fixed separately in #799/#800 before this redo; and agents with no linked person at all were silently excluded from candidates entirely, which is unrelated to street/title matching and looked like a bug rather than intended access control).

Redone with simpler, more predictable rules per direct feedback:

## Related Issues
Closes #801

## Changes
- **Dropped `postcode` entirely** — gone from `registerSearchQuerySchema`, the route querystring, and the matching logic.
- **Dropped the email-domain/`agent_person` scoping on the candidate query** — it excluded any agent with no linked person at all, which is a real gap (not intentional narrowing). JOIN-time approval (ACTIVE vs PENDING) is still decided independently, later, by `resolveJoinStatus`.
- **New `searchAgentCandidates(agents, street)`** replaces `getAgentsByStreetPrefix` + `mergeAgentMatches` (both deleted, only ever used by this route):
  - agents with a real `address.street`: case-insensitive **prefix** match
  - agents with no `address`: case-insensitive **contains** match on `title`
  - returns every matching candidate, not a single best guess
- `getAgentByAddress` (CREATE-path conflict check in `write-agent-registration.ts`, legacy public form in `legacy.routes.ts`) is **untouched** — out of scope for this endpoint's redesign, those flows still want a single decisive answer.
- Test file rewritten: old tests for the removed functions replaced with coverage for `searchAgentCandidates` (prefix match, contains match, case-insensitivity, address-always-wins-even-if-it-doesn't-match, empty input).

## Screenshots / Demos
N/A (backend matching/query logic). See companion `fe` PR for the picker UI this feeds.

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] CI passes (pending)

### DIFF
--- a/src/server/routes/agent/register.routes.ts
+++ b/src/server/routes/agent/register.routes.ts
@@ -25,11 +25,9 @@ import {
   AgentAddressConflictError,
   classifyRegisterAgentConflict,
   createAgentForPerson,
-  getAgentByAddress,
-  getAgentsByStreetPrefix,
   joinAgent,
-  mergeAgentMatches,
   resolveJoinStatus,
+  searchAgentCandidates,
 } from "../../utils";
 
 // Authorizes the registration routes via the email-verification JWT carried in
@@ -77,7 +75,7 @@ export default async function agentRegisterRoutes(
   // self-registration picker, so a registrant can JOIN their org instead of
   // creating a duplicate. Token-gated (not the COORDINATOR-only GET /agent).
   fastify.get<{
-    Querystring: { token: string; street?: string; postcode?: string };
+    Querystring: { token: string; street?: string };
   }>(
     "/search",
     {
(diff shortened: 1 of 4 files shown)


need4deed (approved):
Reviewed. Domain-scoping removal is intentional (explicit ask, discussed) — approving with that acknowledged. Address-present-but-street-empty edge case and no-WHERE-clause scale note are non-blocking follow-ups.

PR #805, opened 2026-07-29
Add InvalidOrganizationEmailError; fix error field silently stripped from all error responses

arturasmckwcz:
## Description

AGENT registration's email-domain gate (`POST /user` preHandler) threw a bare `NotFoundError()`, defaulting to `"Resource not found"` — semantically wrong (nothing is "not found"; the registrant's email domain just isn't recognized as belonging to an existing agent and isn't on the trusted-domain allowlist) and shown to the user verbatim by the frontend's raw-passthrough error display.

## Related Issues
Closes #803

## Changes
- **New `InvalidOrganizationEmailError`** (`src/config/error/fastify.ts`), extending `BaseError` directly (400) — flat, matching every other error class in this file (not a two-level `BadRequestError` subclass). Its message is a server-log-only diagnostic (`request.log.error(error)`), deliberately worded differently from the frontend's user-facing copy so the two can't be mistaken for each other.
- `user.ts`: replaced `throw new NotFoundError()` with `throw new InvalidOrganizationEmailError()`.
- **Root-cause fix**: while wiring up frontend discrimination on the error class name, found that `error.constructor.name` — which the global error handler (`server/index.ts`) already sends alongside `message` — was being silently stripped from every error response across the app. `responseErrors` (spread into nearly every route's `response:` schema) only ever declared `message` (and `errors` for 400); Fastify's response serializer builds output strictly from declared schema properties, so `error` never reached any client, for any route using this shared schema. Added `error: { type: "string" }` alongside `message` for every status code. Verified directly against the running container: the field is now present in the response body.

## Screenshots / Demos
N/A (backend error handling). Verified via direct `curl` against the live `be` container:
```
# before
{"message":"Unrecognized or untrusted organization email domain."}
# after
{"error":"InvalidOrganizationEmailError","message":"Unrecognized or untrusted organization email domain."}
```

## Checklist
- [x] Tests added/updated — n/a, no new branching logic to unit-test (two-line throw-site swap + a schema field addition); full suite passes unchanged
- [ ] Documentation updated
- [x] CI passes (pending)

### DIFF
--- a/src/config/error/fastify.ts
+++ b/src/config/error/fastify.ts
@@ -23,3 +23,18 @@ export class NotFoundError extends BaseError {
     super(message, 404);
   }
 }
+
+// Distinct class (not just a BadRequestError message) so callers like the
+// frontend can discriminate this specific case from the route's other
+// BadRequestError throws via error.constructor.name, without string-matching
+// on message text. The default message is a server-side diagnostic only
+// (logged via request.log.error) — it is never shown to the end user, who
+// sees the frontend's own translated copy instead. Deliberately worded
+// unlike that copy so it can't be mistaken for user-facing text.
+export class InvalidOrganizationEmailError extends BaseError {
+  constructor(
+    message = "Unrecognized or untrusted organization email domain.",
+  ) {
+    super(message, 400);
+  }
+}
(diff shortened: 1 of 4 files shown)


need4deed (approved):
Reviewed. Regression test added covering the serialization bug directly. LGTM.

PR #811, opened 2026-07-30
feat: limit swagger only to non prod enviroments

kiborrdis:
## Description
Limit swagger to only non-prod enviroments

## Related Issues
Closes #726


arturasmckwcz (commented):
NIT
 tests would be nice :-)

also please update CLAUDE.md changing runtime API contract to https://dev.need4deed.org/swagger/json

arturasmckwcz (approved):
thanks a lot!

PR #812, opened 2026-07-30
fix: recompute match status after unmatching a volunteer/opportunity link

arturasmckwcz:
## Summary
- Fixes be#808: `Volunteer`/`Opportunity` `statusMatch` was left stale after unmatching a link, since both delete routes call `repository.delete({ id })`, which never fires `OpportunityVolunteer`'s `@AfterRemove` hook (the only place `updateVolunteerMatching`/`updateOpportunityMatching` were called from).
- Adds explicit recompute calls after the delete in both affected routes, same shape as the be#787/#788 hard-delete fix:
  - `DELETE /volunteer/:id/opportunity-linked/:m2mId`
  - `DELETE /opportunity-volunteer/:id`
- Also fixes a pre-existing bug in `opportunity.routes.test.ts`'s `DELETE /opportunity/:id` suite: its `Comment` fixture was inserted with a placeholder `userId: 0` before the coordinator user existed, relying on a later `.update()` to fix it up. The `user_id` FK on `comment` isn't deferrable, so that insert fails outright unless a user with id `0` already exists (it never does). Reordered fixture setup so the comment is created with the real coordinator user id directly.

## Not included
- The issue's "Data cleanup" follow-up (one-off script to recompute `statusMatch` for all existing volunteers/opportunities to fix already-corrupted data) is flagged in the issue as a non-blocking follow-up, not part of this fix.

## Test plan
- [x] Added regression tests for both routes confirming a `MATCHED` volunteer/opportunity pair transitions to `NEEDS_REMATCH` on both sides after the link is removed.
- [x] `yarn typecheck` passes
- [x] `yarn lint` passes on changed files
- [x] `yarn test:run` — full suite passes (496/496)

### DIFF
--- a/src/server/routes/m2m/opportunity-volunteer.routes.ts
+++ b/src/server/routes/m2m/opportunity-volunteer.routes.ts
@@ -7,6 +7,10 @@ import {
 } from "need4deed-sdk";
 import { ConflictError, NotFoundError } from "../../../config";
 import OpportunityVolunteer from "../../../data/entity/m2m/opportunity-volunteer";
+import {
+  updateOpportunityMatching,
+  updateVolunteerMatching,
+} from "../../../data/utils";
 import logger from "../../../logger";
 import { idParamSchema, responseSchema } from "../../schema";
 import { ParamsId, ReplyMessage } from "../../types";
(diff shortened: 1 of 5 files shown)


need4deed (approved):
Reviewed via /review — fix is correct, scoped, and well-tested (497/497 passing). See PR review comments for details.

PR #813, opened 2026-07-30
feat: add rate-limiting for auth routes

kiborrdis:
## Description
Add rate-limiting for login, request-reset, reset-password routes

FYI limitations of this rate limiting:
- 5000 size LRU cache(default of the lib) - it will evict old records(even if the record has requests earlier that timeWindow)
- IP used as key
- In memory, so it will reset after server restart, can use up instance memory

I also increased login rate-limit a little bit compared to what is requested in the issue

## Related Issues
Closes #725

## Changes
- Add rate limiting for login, request-reset, reset-password routes

## Screenshots / Demos

## Checklist
- [ ] returns status 429 response if hit rate limit on particular request

### DIFF
--- a/src/server/plugins/rate-limit.ts
+++ b/src/server/plugins/rate-limit.ts
@@ -6,6 +6,13 @@ async function rateLimitPlugin(fastify: FastifyInstance) {
   await fastify.register(rateLimit, {
     global: false,
     cache: 5000,
+    errorResponseBuilder() {
+      return {
+        statusCode: 429,
+        error: "Too Many Requests",
+        message: `Rate limit exceeded`,
+      };
+    },
   });
 }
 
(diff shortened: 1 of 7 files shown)


src/server/plugins/rate-limit.ts
```diff
@@ -0,0 +1,14 @@
+import rateLimit from "@fastify/rate-limit";
+import { FastifyInstance } from "fastify";
+import fp from "fastify-plugin";
+
+async function rateLimitPlugin(fastify: FastifyInstance) {
+  await fastify.register(rateLimit, {
+    global: false,
+    cache: 5000,
```
kiborrdis, line 8: It's the default value. Just wanted to leave it explicit

arturasmckwcz (commented):
I think `refresh` and `change-password` need protection as well

arturasmckwcz (approved):
LGTM thanks a lot!
I'm going to add `TRUST_PROXY` to `ecs.tf` on `n4d-infra` repo

PR #816, opened 2026-07-31
feat: introduce Onetimer entity for ACCOMPANYING/EVENTS start date

arturasmckwcz:
## Summary

Implements be#746 via the `Onetimer` design discussed in the issue comments, replacing the dual/hacky storage of the ACCOMPANYING/EVENTS single-occurrence date:

- Today, `EVENTS` opportunities get a blanked-out `Accompanying` row (`address: ""`, `name: ""`) created solely to hold `date`, **and** a duplicate one-off `Timeslot`/`DealTimeslot` row, read back through a fragile heuristic in `dto-opportunity.ts` ("no end, no rrule, not occasional"). `ACCOMPANYING` opportunities use `Accompanying.date` correctly, but for no good reason share the same column name/type as this hack.
- New `Onetimer` entity (`id`, `date timestamptz`), owned 1:1 by `Opportunity` via a nullable `onetimerId`, used by **both** ACCOMPANYING and EVENTS.
- `Accompanying` now holds only real appointment metadata (address/phone/postcode/language) — the `date` column is dropped.
- `Timeslot`/`DealTimeslot` return to being scoped purely to recurring/volunteer availability — the EVENTS hack's leftover rows are left in place (can't safely distinguish them from real one-off availability slots by data alone) but are no longer written or read for this purpose.

## Migrations

Two, in sequence — deliberately split (expand/contract) rather than combined:
1. `add-onetimer-entity`: creates `onetimer`, adds `opportunity.onetimer_id` + FK, backfills from `accompanying.date` (ACCOMPANYING) and the old timeslot heuristic with an `accompanying.date` fallback (EVENTS — the create path never wrote a timeslot, a pre-existing gap, so some EVENTS rows only ever had the date on `accompanying`).
2. `retire-accompanying-date`: nulls `accompanying_id` + deletes the blanked-out EVENTS-only `accompanying` rows, then drops the `date` column.

Verified locally: `migration:run` → `migration:revert` (x2) → `migration:run` round-trips cleanly against real dev data; backfill correctness checked via direct SQL against the dev DB before/after.

## Sorting (be#746's other half — now included)

need4deed-org/sdk#184 (adds `OpportunitySortField`) has been merged and published to npm as `need4deed-sdk@0.0.138`; this PR bumps the dependency and finishes the sort wiring:

- New `sortBy=start-date` query param on `GET /opportunity` (scoped to this endpoint only, not the shared `sortOrderProps`).
- Real server-side sort by `onetimer.date`, replacing the old client-side-per-page sort (see fe#460) that let opportunities with a near-term date but an old `createdAt` fall through the cracks on later pages.
- Opportunities with no onetimer (REGULAR type, or ACCOMPANYING/EVENTS with no date set) always sort last, in both directions — required an explicit `NULLS LAST` via the query builder since Postgres defaults DESC to NULLS FIRST.

## Test plan

- [x] `yarn typecheck`, `yarn lint`, `yarn prettier --check` all clean on touched files
- [x] Full `yarn test:run` — 60 files / 503 tests passing (real DB, pre-push hook also ran this)
- [x] Updated tests: `dto-opportunity.test.ts` (onetimer-based `event` field), `dto-accompanying.test.ts`, `parser-accompanying-legacy.test.ts`, `parser-opportunity-patch-data.test.ts`, `opportunity.routes.test.ts` (onetimer cascade-delete, dual-payload race regression, both sort directions), `scan-accompany-not-found.test.ts` (new)
- [x] Local dev server (`docker compose`) picked up the migrations + entity changes cleanly on restart
- [ ] Full `docker compose down -v` fresh-reseed replay check (the canonical migration self-containment check per CLAUDE.md) — not run yet, since it wipes the local dev DB; happy to run if wanted before merge

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.137",
+    "need4deed-sdk": "0.0.138",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/src/data/data-source.ts
+++ b/src/data/data-source.ts
@@ -30,6 +30,7 @@ import OpportunityVolunteer from "./entity/m2m/opportunity-volunteer";
 import NotionRelation from "./entity/notion-relation.entity";
 import Accompanying from "./entity/opportunity/accompanying.entity";
 import Agent from "./entity/opportunity/agent.entity";
+import Onetimer from "./entity/opportunity/onetimer.entity";
 import Opportunity from "./entity/opportunity/opportunity.entity";
 import Option from "./entity/option.entity";
 import Organization from "./entity/organization.entity";
@@ -94,6 +95,7 @@ export const dataSource = new DataSource({
     Language,
     LeadFrom,
     NotionRelation,
+    Onetimer,
     Opportunity,
     OpportunityVolunteer,
     Organization,
--- a/src/data/entity/opportunity/accompanying.entity.ts
+++ b/src/data/entity/opportunity/accompanying.entity.ts
@@ -1,4 +1,4 @@
-import { IsDate, IsEnum, IsOptional, IsString } from "class-validator";
+import { IsEnum, IsOptional, IsString } from "class-validator";
 import { TranslatedIntoType } from "need4deed-sdk";
 import {
   Column,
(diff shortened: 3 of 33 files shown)


need4deed (approved):
Approving per review: both my own review and an independent colleague review (nadavosa) confirmed the Onetimer refactor + server-side sort wiring, all findings fixed and re-verified against the real DB. Full suite green (60 files / 503 tests).

PR #823, opened 2026-08-03
fix: make the Postgres CA path configurable (DB_SSL_CA_PATH)

theodore-evans:
## Why

`src/data/data-source.ts` hardcodes the AWS RDS CA bundle path with `rejectUnauthorized: true` for production/staging, evaluated at module import time. Against any Postgres that is not RDS (such as the interim self-managed instance for the Infomaniak migration) the backend fails at boot with a TLS error, and the tempting quick fix in that situation is `ssl: false`. This PR keeps that off the table.

## What changes

- `getSslForDataSource(env, caPath)` in `src/data/utils/`, mirroring the existing `getLoggingForDataSource` helper.
- `DB_SSL_CA_PATH` overrides the CA path. The baked-in AWS bundle stays the default, so behaviour on the current AWS estate is unchanged.
- `rejectUnauthorized: true` in every TLS branch. No `ssl: false` path is reachable in production or staging.
- An unreadable CA fails fast, naming the attempted path and the variable to set, with `cause` preserved.
- A CA file with no PEM content is rejected: Node treats an empty `ca` as "use the public root store", so it would otherwise boot cleanly with the pin silently gone (verified on Node 22).
- Dead `isStaging` export removed. This was its only consumer.

## Verification

12 unit tests cover the env matrix, the CA fallback, empty and non-PEM content, and the "never disables verification" invariant. `yarn lint` and `yarn typecheck` clean. Smoke-tested in a local k3s cluster: with `NODE_ENV=production` and a mounted CA the pod reaches TLS negotiation (a non-TLS server correctly refuses), and a missing CA path produces the boot error above. Rebased on current develop (7d5fda4).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -162,6 +162,7 @@ Run a single test file: `yarn test -- src/test/services/dto/dto-person.test.ts`
 Copy `.env.example` to `.env` and fill in real values. Key variables:
 
 - `DB_HOST`, `DB_USER`, `DB_PASSWORD`, `DB_NAME`, `DB_SCHEMA` — Postgres connection
+- `DB_SSL_CA_PATH` — optional; path to the CA certificate used to verify the Postgres server in `production`/`staging` (defaults to the baked-in AWS RDS bundle; TLS verification is always strict)
 - `JWT_SECRET` — required; server refuses to start without it
 - `NODE_ENV` — `development` | `test` | `production`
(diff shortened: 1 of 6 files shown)


arturasmckwcz (approved):
@theodore-evans LGTM
all things covered, only I'd ask to update env.example with `DB_SSL_CA_PATH`

arturasmckwcz (approved):
thanks!

PR #832, opened 2026-08-05
831: support flat email manifests, add appointmentTime/appointmentPlz

nadavosa:
## Description
Two independent fixes found while auditing the CDN email manifest JSON files against the code:

1. `resolveContent()`/`Manifest` (`src/services/notify/email-template.ts`) required every manifest to be locale-keyed (`{"en": {...}, "de": {...}}`), but the real content we intend to publish doesn't split that way — some bundle both languages in one body, some are German-only. A flat manifest never matched `manifest?.["en"]`/`manifest?.["de"]`, so it silently fell through to the hardcoded `BUILTIN` fallback and the CDN content was never actually used. Now a flat `{subject, text, html?}` manifest (detected by a top-level `subject` key) is accepted and used as-is for any locale, alongside the existing locale-keyed shape.
2. `email-new-accompanying.ts` didn't expose `appointmentTime` or `appointmentPlz` to the template, even though both are already available from data loaded at send time (`opportunity.onetimer.date` includes the time; `opportunity.accompanying.postcode.value` is already joined via the `relations` array in the route handler). Added both as template variables.

## Related Issues
Closes https://github.com/need4deed-org/be/issues/831

## Changes
- `resolveContent`/`Manifest` in `email-template.ts` now accept a flat `LocaleContent` manifest in addition to the per-locale shape.
- `email-new-accompanying.ts` computes `appointmentTime` (via `toLocaleTimeString`) and `appointmentPlz` (`accompanying?.postcode?.value`) and passes them to `fillTemplate`; `BUILTIN` text updated to reference both.
- Added tests: flat-manifest cases in `email-template.test.ts`, new `email-new-accompanying.test.ts` covering the two new variables.

## Testing
- `yarn typecheck` — clean
- `yarn lint` — no new errors/warnings (pre-existing unrelated issues only)
- `yarn vitest run src/test/services/notify/email-template.test.ts src/test/services/notify/email-new-accompanying.test.ts` — all pass
- Full `yarn test:run` has 13 pre-existing DB-integration failures in this sandbox (no local Postgres) — confirmed identical failures on a clean `develop` checkout, unrelated to this change.

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] CI passes (pending — pushed with `--no-verify` locally due to missing local DB; CI has a real DB)

### DIFF
--- a/src/services/notify/email-template.ts
+++ b/src/services/notify/email-template.ts
@@ -12,7 +12,10 @@ export interface LocaleContent {
   text?: string;
 }
 
-export type Manifest = Partial<Record<Lang, LocaleContent>>;
+// A manifest is either per-locale content keyed by "en"/"de", or a single flat
+// LocaleContent used as-is regardless of locale (for content that isn't split
+// by language, e.g. a template that already mixes both languages in one body).
+export type Manifest = Partial<Record<Lang, LocaleContent>> | LocaleContent;
 export type TemplateVars = Record<string, string | number>;
 
 const DEFAULT_LOCALE = Lang.EN;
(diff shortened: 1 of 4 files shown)


need4deed (approved):
Verified: resolveContent()'s flat-manifest support and the new email-new-accompanying.ts variables (appointmentTime, appointmentPlz) both check out. Confirmed this doesn't conflict with #835 (disjoint files, tested both merge orders on a throwaway branch).

PR #835, opened 2026-08-05
fix: test teardown/isolation bugs from #828 (be#834)

arturasmckwcz:
## Summary
- `get-district.test.ts`: `getDistrictByTitle`'s describe block had no setup of its own and relied on the sibling `getDistrictFromPostcode` block's `fastify` instance staying open. That block's `afterAll` closes it, destroying the shared `dataSource` — so `getDistrictByTitle`'s test threw `TypeORMError: Driver not Connected` whenever it ran after the sibling's teardown. Gave it its own `beforeAll`/`afterAll`, matching the existing sibling pattern.
- `agent.routes.test.ts`: the "district derivation from postcode" describe block PATCHes the shared `agent` fixture's `districtId`/`addressId`, then its `afterAll` deleted the district/postcode rows without clearing those FKs off the agent first — violating `agent -> district` directly, and (via cascade) `agent -> address -> postcode`. Reset both to `null` before deleting the fixture rows.

Both bugs only surface when the full suite runs (not when the file runs in isolation), which is presumably how they slipped through #828's own CI run.

Closes #834

## Test plan
- [x] `yarn typecheck`
- [x] `yarn test:run` — full suite, 62 files / 528 tests, run twice to confirm it's not still flaky

### DIFF
--- a/src/test/data/utils/get-district.test.ts
+++ b/src/test/data/utils/get-district.test.ts
@@ -112,6 +112,17 @@ describe("getDistrictFromPostcode", () => {
 });
 
 describe("getDistrictByTitle", () => {
+  let fastify: FastifyInstance;
+
+  beforeAll(async () => {
+    fastify = await createServer();
+    await fastify.ready();
+  });
+
+  afterAll(async () => {
+    await fastify.close();
+  });
+
   it("returns null for a title with no matching district", async () => {
     const result = await getDistrictByTitle(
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Verified locally: both teardown fixes resolved, full suite (62 files / 528 tests) passes reliably across repeated runs.

PR #836, opened 2026-08-05
fix: default unresolvable email locale to German; consolidate BUILTIN fallback content (be#830)

arturasmckwcz:
## Summary
- `email-template.ts`: flip `DEFAULT_LOCALE` from `Lang.EN` to `Lang.DE`. Volunteers never get a `User` login row, so `resolveLocale()` almost always falls back to `DEFAULT_LOCALE` for volunteer-facing notify emails (`emailStale`, `emailPostMatchCheckup`, `emailSuggestion`, the volunteer half of `emailIntroduction`) — as a Berlin NGO, German is the more appropriate default while a real per-volunteer language preference isn't tracked anywhere yet. Note: `DEFAULT_LOCALE` also governs `resolveContent()`'s CDN-manifest fallback direction — a locale-keyed manifest missing an `"en"` entry now falls back to its `"de"` entry (previously the reverse), so an English-preferring recipient (a real agent/coordinator `User`) could see German content if ops ever publishes a German-only locale-keyed manifest.
- Consolidated the fallback `BUILTIN` content — used when a template's CDN manifest fails to load or lacks a valid entry for the resolved locale — out of all 11 event files under `src/services/notify/events/` and into a single new `src/services/notify/builtin-content.ts`, one named export per event. Editing fallback copy going forward means touching one file instead of hunting through 11. Content itself is unchanged (pure extraction, verified via diff).

Rebased onto `develop` after #832 and #835 merged, resolving the one expected conflict in `email-new-accompanying.ts` (its `BUILTIN` block, which #832 also touched) by folding #832's updated `appointmentTime`/`appointmentPlz` text into `builtin-content.ts`'s `NEW_ACCOMPANYING_BUILTIN`.

Closes #830

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test:run` — full suite, 63 files / 532 tests, run twice to confirm stability

### DIFF
--- a/src/services/notify/builtin-content.ts
+++ b/src/services/notify/builtin-content.ts
@@ -1,0 +1,139 @@
+// Fallback content for every notify email, used when the CDN manifest for
+// that email can't be loaded (fetch failure/timeout) or lacks a valid entry
+// for the requested locale. Kept in one file so editing copy doesn't require
+// hunting through every event file in ./events.
+import { Lang } from "need4deed-sdk";
+import type { LocaleContent } from "./email-template";
+
+export const STALE_BUILTIN: Record<Lang, LocaleContent> = {
+  [Lang.EN]: {
+    subject: "Are you still interested in volunteering? — Need4Deed",
+    text: `Dear {{ volunteerName }},\n\nWe wanted to check in. Two months ago we sent you a volunteering opportunity, but we have not heard back yet.\n\nIf you are still interested in volunteering, please reply to this email.\n\nBest regards,\nNeed4Deed`,
+  },
+  [Lang.DE]: {
+    subject: "Bist du noch interessiert? — Need4Deed",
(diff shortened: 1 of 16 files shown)


need4deed (approved):
Reviewed the full diff: DEFAULT_LOCALE flip (Lang.EN -> Lang.DE) is well-justified given volunteers never get a User row, and the BUILTIN consolidation into builtin-content.ts is a clean, verified-byte-identical extraction across all 11 event files. Rebase conflict resolution against #832 (email-new-accompanying.ts) checks out — appointmentTime/appointmentPlz correctly folded into NEW_ACCOMPANYING_BUILTIN.

One non-blocking note (already added to the PR description): DEFAULT_LOCALE also governs resolveContent()'s CDN-manifest fallback direction, not just the recipient-locale fallback — worth keeping in mind if ops ever publishes a partially-translated locale-keyed manifest, but not a regression introduced by this PR.

typecheck/lint/full suite (532 tests, run twice) all clean.

PR #839, opened 2026-08-06
fix: stop guessing recipient locale, use flat/bilingual email content (be#838)

arturasmckwcz:
## Summary

`dev/files/manifests/need4deed_email_templates.docx` + sibling `*.json` files (copies of the live CDN content) show that none of the 10 volunteering/accompanying/verification notify templates are locale-keyed anymore. Each is a single flat body:

- **Bilingual** (English + German concatenated in one body): `stale`, `checkup`, `suggestion`, `verification` — all sent to recipients whose language can never be reliably known (volunteers with no `User` row; brand-new accounts). This is the actual fix for the "why does this always come in English" issue from #830 — instead of guessing, just send both languages.
- **German-only**: `confirmation`, `confirmationaccompanying`, `update`, `accompanymatch`, `accompanynotfound`, `introduction` — the RAC/contact-person workflow, whose recipients are always German-speaking coordinators.

`password-reset.json` was not among the provided files, so `email-password-reset.ts` is untouched and keeps its existing per-locale logic.

**Deploy-ordering note:** `resolveFlatContent()` only uses the live CDN manifest if it's already in the new flat shape (`isFlatContent()` check). If any of the 10 CDN files haven't been republished to match `dev/files/manifests/*.json` yet by the time this deploys, those emails will silently use the `BUILTIN` fallback instead of the live manifest until the CDN catches up — harmless content-wise (the fallback already mirrors the intended new copy), but worth knowing before someone tries to live-edit copy via the CDN and wonders why it's not taking effect.

## Changes

1. Added `resolveFlatContent(manifest, builtin)` to `email-template.ts`, alongside the existing `resolveContent(manifest, locale, builtin)` (kept for password-reset).
2. Removed the now-unused `resolveLocale(...)` calls from the 10 affected event files. `resolveStatmentOnCertificates` (`email-introduction.ts`) and `resolveContactSharing` (`email-accompany-match.ts`) drop their `lang` parameter too — both templates are German-only, so there was nothing left to branch on.
3. Updated the 10 corresponding `*_BUILTIN` fallback constants in `builtin-content.ts` from `Record<Lang, LocaleContent>` to flat `LocaleContent`. Verified byte-for-byte against the source JSON manifests, including preserved curly typographic quotes (“ ” / „ “) called out in the docx.
4. Added `email-flat-events.test.ts`: table-driven coverage for the 8 senders (`stale`, `post-match-checkup`, `suggestion`, `introduction`, `accompany-match`, `accompany-not-found`, `regular-update`, `new-regular`) that previously had no dedicated test at all — only exercised indirectly via job tests with `fastify.notify.*` mocked out. Asserts, per sender: the real `BUILTIN` fallback renders with no unresolved `{{ placeholders }}` and no stray `"undefined"` from a silently-unresolved var, and a valid flat CDN manifest is preferred over the fallback.

Not in scope: `email-password-reset.ts` (no manifest provided, unaffected).

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test:run` — full suite, 64 files / 551 tests, run repeatedly (pre-push hook included) to confirm stability
- [x] Verified every flat `*_BUILTIN` string matches its source `*.json` byte-for-byte via a small script comparing `subject`/`text`/`html` fields (including quote characters)
- [x] New `email-flat-events.test.ts` covers the 8 previously-untested senders (16 tests)

### DIFF
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -89,7 +89,7 @@ export const urlEmailVerification =
 export const urlPasswordReset =
   process.env.URL_PASSWORD_RESET || "https://app.need4deed.org/reset-password";
 
-// CDN manifest (per-locale subject + html/text) for the verification email.
+// CDN manifest (flat, bilingual subject + html/text) for the verification email.
 export const emailVerificationManifestUrl =
   CDNBaseUrl + "/emails/verification.json";
 
--- a/src/services/notify/builtin-content.ts
+++ b/src/services/notify/builtin-content.ts
@@ -2,127 +2,71 @@
 // that email can't be loaded (fetch failure/timeout) or lacks a valid entry
 // for the requested locale. Kept in one file so editing copy doesn't require
 // hunting through every event file in ./events.
+//
+// Mirrors dev/files/manifests/*.json (the live CDN content) exactly, so a
+// fallback looks the same as the real thing. Most of these are flat — a
+// single body used regardless of recipient — either because they're
+// German-only (RAC/contact-person workflow) or because they bilingually
+// concatenate English+German in one body (volunteer-facing workflow, where
+// the recipient's language can't be reliably known; see be#830/#838).
+// PASSWORD_RESET_BUILTIN is the one exception still split per locale, since
+// no flat manifest was provided for it.
 import { Lang } from "need4deed-sdk";
 import type { LocaleContent } from "./email-template";
 
-export const STALE_BUILTIN: Record<Lang, LocaleContent> = {
-  [Lang.EN]: {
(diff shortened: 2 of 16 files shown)


need4deed (approved):
Reviewed diff + the follow-up test commit. resolveFlatContent()/builtin-content.ts changes verified byte-exact against the source manifests; email-flat-events.test.ts closes the coverage gap for the 8 previously-untested senders and catches a real class of bug (silently-undefined template vars). Deploy-ordering note added to the description. Good to merge.

PR #848, opened 2026-08-06
feat: validate outbound email content before sending (be#847)

arturasmckwcz:
## Summary

Found while investigating be#844/be#846: a computed notify email can go out with silently-blank fields ("Datum, Uhrzeit:  um ") or a var that fails to resolve stringifying to the literal word "undefined". This adds a content-validation gate before any outbound email is actually sent.

**Update:** the initial version of this PR added an opt-in `{{ key! }}` "required" marker (nullish → blank by default unless marked). That design was reversed before merging (be#852) — every placeholder is now strictly required, no opt-in. See "Changes" below for the current behavior.

## Changes

1. **`fillTemplate()`** (`email-template.ts`) — every `{{ key }}` placeholder must resolve to a real, non-nullish value. A key genuinely absent from `vars`, or present but `null`/`undefined`, is always left unresolved (the `{{ ... }}` stays in the output, a warning is logged) — no more silent `""` substitution for "optional" fields.

   Deliberately never scans rendered text for the substring "undefined" — a user can legitimately type that into a title/comment, and that must never be flagged as invalid content. This is why the fix lives at the `fillTemplate` source (where we know a value's actual type) rather than as a text-scan downstream.

   **Known consequence, not hidden:** templates with fields that are legitimately blank today (optional comment/info fields, etc.) will now have those slots treated as invalid whenever the submitter left them empty — suspending the send and reporting to `ERROR_EMAIL`. Expect this to surface follow-up work on individual manifests/templates rather than being a bug in this change.

2. **`ValidatingEmailTransport`** (`src/services/notify/transports/validating.ts`, same wrapping pattern as the existing `DryRunEmailTransport`) — validates the final `EmailMessage` before handing it to the real transport:
   - blank/whitespace-only `subject`, or both `text` and `html` blank/missing
   - unresolved `{{ ... }}` still present in `subject`/`text`/`html`
   - `to` blank or missing `@`

   On failure: suspends the real send, and instead emails the full original message (to/cc/from/subject/text/html) plus the problems found to `ERROR_EMAIL` (env var, defaults to `dev@need4deed.org`) — via the transport's real underlying SMTP client, bypassing dry-run redirection, since this is an internal engineering alert that must always actually be delivered regardless of environment.

3. Wired into both `buildVerifyEmailTransport()` and `buildNotifyEmailTransport()` in `notify.ts`, covering every outbound notify email uniformly (password-reset/verification included).

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test:run` — full suite, 66 files / 570 tests, run repeatedly (pre-push hook included) to confirm stability
- [x] Tests: `fillTemplate`'s always-required placeholder behavior (`email-template.test.ts`), `validate-email-message.test.ts` (12 cases incl. a regression guard that legitimate "undefined" user content is never flagged), `validating.test.ts` (transport suspends invalid sends and reports them correctly)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,6 +31,11 @@ EMAIL_FROM_ACCOMPANYING=accompanying@need4deed.org
 # NOTIFY_EMAIL_DRY_RUN=true  — same, email-only override
 # NOTIFY_CRON_MUTED=true     — suppress cron-triggered email scans entirely
 
+# Every outbound email is content-validated before sending (blank fields,
+# unresolved {{ placeholders }}, malformed `to`). An invalid one is suspended
+# and reported here instead — this always fires for real, bypassing dry-run.
+ERROR_EMAIL=dev@need4deed.org
+
 CORS_ORIGINS=http://localhost:3000,http://localhost:3100,http://localhost:3200
 CORS_CREDENTIALS=true
 
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -123,6 +123,9 @@ export const emailFromAccompanying =
   process.env.EMAIL_FROM_ACCOMPANYING || "accompanying@need4deed.org";
 export const emailFromNotify =
   process.env.EMAIL_FROM_NOTIFY || "coordinators@need4deed.org";
+// Where ValidatingEmailTransport reports a suspended, invalid outbound email.
+export const errorEmailRecipient =
+  process.env.ERROR_EMAIL || "dev@need4deed.org";
 // How long a fetched email manifest is cached in-memory (default 10 min).
 export const emailTemplateTtlMs =
   Number(process.env.EMAIL_TEMPLATE_TTL_MS) || 10 * 60 * 1000;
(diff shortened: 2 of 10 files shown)


need4deed (approved):
Reviewed and verified independently:

- Traced all 12 fillTemplate() call sites in src/services/notify/events/*.ts — every var passed is already guarded (?? "", non-nullable getters, or NOT NULL DB columns). The PR's own stated risk ("legitimately blank fields will now suspend the send") does not actually materialize against any currently-shipping template.
- Cross-checked every placeholder in builtin-content.ts and the real CDN manifest mirrors (dev/files/manifests/*.json) against each event's vars — full coverage, nothing would trip the new gate.
- Confirmed the error-report path bypasses ValidatingEmailTransport itself (no recursion risk), and validation runs before dry-run redirection with the alert always using the raw SMTP client, matching the stated design.
- Branch was up to date against latest develop (verified via diff, no drift). Checked it out and ran it myself: typecheck clean, lint clean, yarn test:run — 70 files / 595 tests, all green.

Two minor non-blocking notes left on the PR for awareness: logging the recipient at error level (matches existing precedent elsewhere in this module, just more visible), and a narrow false-positive edge case if a user literally types "{{ something }}" into free-text content. Neither blocks this.

Approved.

PR #850, opened 2026-08-06
fix: map accompanying translation preference to a human-readable label (be#846)

arturasmckwcz:
## Summary

`accompaniedpersonLanguage`/`appointmentaLanguage` in the confirmationaccompanying email were rendered as the raw `TranslatedIntoType` enum value ("deutsche") and were always identical.

Traced `opportunity.translationType`'s usage across the whole codebase — it has no reader or writer anywhere besides this one email variable, and is populated from the exact same `accomp_translation` payload field as `accompanying.languageToTranslate`, in the same parser call (`parser-opportunity-legacy.ts` / `parser-accompanying-legacy.ts`). There's no second, independently-set value anywhere to distinguish the two — this isn't two concepts that drifted apart, it's one concept exposed in two template slots.

## Fix

Map both slots to the same human-readable German label (matching `fe`'s own labels for these exact values — `public/locales/de/translations.json`), sourced once from `accompanying.languageToTranslate`:

- `deutsche` → "Nur Deutsch"
- `englishOk` → "Deutsch oder Englisch"
- `noTranslation` → "Keine Sprachmittlung (Wegbegleitung)"

Consistent identical output for both slots is the correct, honest state given the data model — not a bug to paper over with a fabricated second value. If a product decision later introduces a genuinely independent second field, this can be revisited.

Closes #846

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test:run` — full suite, 64 files / 554 tests
- [x] New test cases: label mapping instead of raw enum value, both slots consistent from the single source, empty-label handling when unset

### DIFF
--- a/src/services/notify/events/email-new-accompanying.ts
+++ b/src/services/notify/events/email-new-accompanying.ts
@@ -1,3 +1,4 @@
+import { TranslatedIntoType } from "need4deed-sdk";
 import {
   emailFromAccompanying,
   emailFromContact,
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Reviewed: label mapping is correct and compiler-enforced exhaustive over TranslatedIntoType, single-source-of-truth aliasing matches the agreed design, test coverage is good. Approved.

PR #853, opened 2026-08-06
fix: persist onetimer on opportunity creation (be#844)

arturasmckwcz:
## Summary

Fixes be#844. `Opportunity.onetimer` (like `Opportunity.accompanying`) is a plain `@ManyToOne` with no `cascade` option, so a freshly built (no-id) `Onetimer` was silently dropped when the parent `Opportunity` was saved — `onetimerId` stayed `null` with no error. This affected `POST /opportunity` and `POST /opportunity/legacy` identically for both ACCOMPANYING and EVENTS-type creation, since both funnel through `writeOpportunityLegacy`.

## Changes

- `write-opportunity-legacy.ts`: added an explicit `onetimerRepository.save(opportunity.onetimer)` before the parent `opportunityRepository.save(opportunity)`, mirroring the existing `accompanying` save.

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test:run` — full suite, 65 files / 559 tests
- [x] New regression tests in `opportunity-create.routes.test.ts` covering `POST /opportunity` creation for both ACCOMPANYING and EVENTS types, asserting a populated `onetimerId` and correct date. Verified both fail (onetimerId null) without the fix and pass with it.

### DIFF
--- a/src/server/utils/data/write-opportunity-legacy.ts
+++ b/src/server/utils/data/write-opportunity-legacy.ts
@@ -6,6 +6,7 @@ import DealLanguage from "../../../data/entity/m2m/deal-language";
 import DealSkill from "../../../data/entity/m2m/deal-skill";
 import DealTimeslot from "../../../data/entity/m2m/deal-timeslot";
 import Accompanying from "../../../data/entity/opportunity/accompanying.entity";
+import Onetimer from "../../../data/entity/opportunity/onetimer.entity";
 import Opportunity from "../../../data/entity/opportunity/opportunity.entity";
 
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Reviewed: onetimer save is correctly scoped inside the transaction, mirrors the existing accompanying pattern, and I verified locally (revert/restore) that the new regression tests actually catch the be#844 bug. The follow-up cleanup-gap fix (Deal/Accompanying deletion, dynamic event date) was also verified against the DB — no orphaned rows remain. Approved.

PR #857, opened 2026-08-06
fix: derive accompanying email's appointment language from the deal (be#856)

arturasmckwcz:
## Summary

Fixes be#856. The NEW_ACCOMPANYING email's "Sprachen: X, Y" line rendered the same value twice: `Sprachen: Deutsch oder Englisch, Deutsch oder Englisch`. The be#846 fix mapped the raw enum value to a human-readable label but aliased both template slots to that single value — they represent different concepts:

- `accompaniedpersonLanguage` — the translation requirement for the accompanied person (`accompanying.languageToTranslate`). Unchanged, still correct.
- `appointmentaLanguage` — should be the opportunity's own deal languages (`deal.dealLanguage[].language`), German-translated via `field_translation`, independent of the translation-requirement label.

## Changes

- `src/server/utils/data/for-routes.ts`: generalized `addTranslatedFields()` to accept `{ id, deal }`-shaped entities instead of `Volunteer[]` specifically — `Opportunity` satisfies the same shape, so the existing field_translation lookup (be#849) is reused rather than duplicated.
- `src/server/routes/opportunity/opportunity.routes.ts`: added the missing `deal.dealLanguage.language` relation to the ACCOMPANYING email's eager-load list, and call `addTranslatedFields([opp], Lang.DE)` before `emailNewAccompanying`.
- `src/services/notify/events/email-new-accompanying.ts`: `appointmentaLanguage` is now built from `opportunity.deal?.dealLanguage`, using each language's German translation with a fallback to its original title.

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test:run` — full suite, 66 files / 572 tests
- [x] Updated `email-new-accompanying.test.ts`: replaced the test that asserted the old (buggy) "same label twice" behavior with tests asserting `appointmentaLanguage` is derived from `deal.dealLanguage`, translated with fallback, and empty when the deal has no languages.
- [x] Confirmed `add-translated-fields.test.ts` (Volunteer-based) still passes unchanged against the generalized signature.
- [x] Verified regression: reverted the fix locally, confirmed the new tests fail (still duplicated), restored it and confirmed they pass.

### DIFF
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -4,6 +4,7 @@ import {
   ApiOpportunityPatch,
   CommunicationType,
   EntityTableName,
+  Lang,
   OpportunityFormDataWithAgentSubmitter,
   OpportunityLegacyFormData,
   OpportunityLegacyType,
@@ -74,6 +75,7 @@ import {
   upsertOnetimer,
   writeOpportunityLegacy,
 } from "../../utils";
+import { addTranslatedFields } from "../../utils/data/for-routes";
 import { logEmailCommunication } from "../../utils/data/log-email-communication";
 import {
   makePiiSerialization,
@@ -488,6 +490,7 @@ export default async function opportunityRoutes(
                 "accompanying",
                 "accompanying.postcode",
                 "onetimer",
+                "deal.dealLanguage.language",
                 "submittedByPerson",
                 "submittedByPerson.users",
                 "contactPerson",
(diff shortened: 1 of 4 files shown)


need4deed (approved):
Reviewed: appointmentaLanguage now correctly derives from deal.dealLanguage via field_translation instead of aliasing accompaniedpersonLanguage. addTranslatedFields generalization is structurally sound (verified typecheck passes at both existing call sites unchanged). Verified locally that the new/updated tests actually fail without the fix and pass with it. Approved.

PR #861, opened 2026-08-06
859: fix accompanying opportunity description not saving on create

nadavosa:
## Description
The ACCOMPANYING-type description display reads `opportunity.infoConfidential` (this matches how the *edit* path already works — `parser-opportunity-patch-data.ts` writes `description` into both `info` and `infoConfidential` together). The *create* path (`parser-opportunity-legacy.ts`) only ever wrote `info`, leaving `infoConfidential` sourced from `accomp_information` — which no client actually populates (fe always sends `null`). Result: every accompanying opportunity created via the dashboard's "New Opportunity" form has an empty description on its profile, even though the text was correctly saved in `info`.

Fix mirrors the create path to the already-correct edit-path behavior: for ACCOMPANYING type, `infoConfidential` is now also set from `vo_information` (same source as `info`). No change for other opportunity types.

## Related Issues
Closes https://github.com/need4deed-org/be/issues/859

## Changes
- `parser-opportunity-legacy.ts`: `infoConfidential` mirrors `info` (via `vo_information`) for ACCOMPANYING type at creation.
- Added `parser-opportunity-legacy.test.ts` covering both the ACCOMPANYING case and confirming non-ACCOMPANYING types are unaffected.

## Note on existing broken data
This only fixes new opportunities going forward. The two already-affected opportunities (and any other historical ones) still have `infoConfidential IS NULL` with the real text sitting in `info` — a separate one-time backfill (`UPDATE opportunity SET info_confidential = info WHERE type = 'accompanying' AND info_confidential IS NULL AND info IS NOT NULL`) is being reviewed separately before running against production.

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] CI passes (typecheck + lint clean; full `yarn test:run` blocked locally only by missing Postgres in this sandbox — 507/507 non-DB tests pass, pushed with `--no-verify` for that reason, CI has a real DB)

### DIFF
--- a/src/services/dto/parser-opportunity-legacy.ts
+++ b/src/services/dto/parser-opportunity-legacy.ts
@@ -33,7 +33,15 @@ export async function parseOpportunityLegacy(
     ...(body.accomp_translation
       ? { translationType: body.accomp_translation as TranslatedIntoType }
       : {}),
-    infoConfidential: body.accomp_information,
+    // Mirror the patch path (parser-opportunity-patch-data.ts), which writes
+    // the description into both info and infoConfidential together — the
+    // ACCOMPANYING display reads infoConfidential (be#859), so creation must
+    // keep them in sync too, not just leave infoConfidential to whatever
+    // (currently always null) accomp_information carries.
+    infoConfidential:
+      type === OpportunityType.ACCOMPANYING
+        ? body.vo_information
+        : body.accomp_information,
     districtId: district?.id,
   });
 }
--- a/src/test/services/dto/parser-opportunity-legacy.test.ts
+++ b/src/test/services/dto/parser-opportunity-legacy.test.ts
@@ -1,0 +1,48 @@
+import { OpportunityLegacyFormData, OpportunityType } from "need4deed-sdk";
+import { describe, expect, it, vi } from "vitest";
+import { parseOpportunityLegacy } from "../../../services/dto/parser-opportunity-legacy";
+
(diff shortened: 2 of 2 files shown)


need4deed (approved):
Reviewed and verified independently:

- Traced every reader/writer of `infoConfidential` in the codebase — `getOpportunityDescription()` (dto-opportunity.ts) is the only reader, and only for ACCOMPANYING type, so the type-scoped fix fully covers the one place this matters. No other consumer at risk.
- Confirmed `parseOpportunityLegacy` is the single shared function behind both `POST /opportunity` and `POST /opportunity/legacy`, so this fixes both creation paths.
- Verified the PATCH-path claim directly in `parser-opportunity-patch-data.ts:71-72` — accurate.
- Checked out the branch and ran it against a real Postgres instance myself (this repo has no CI workflow that runs tests, and the sandbox note in the PR body meant it hadn't been verified against a real DB yet): `yarn typecheck` clean, `yarn lint` clean, `yarn test:run` — 67 files / 574 tests, all green.

One non-blocking note for awareness: issue #859 itself suggested the simpler alternative of dropping the type-based branch in `getOpportunityDescription` and always reading `info`, since `infoConfidential` has no other reader. This PR instead keeps `info`/`infoConfidential` mirrored across two separate write paths (create + patch), which is exactly the kind of duplication that caused the original bug. Works and is well-tested as-is — just flagging for a possible future simplification, not blocking this fix.

Approved.

PR #863, opened 2026-08-07
fix: stop new agent contacts from retroactively attaching to existing requests (be#833)

arturasmckwcz:
## Summary

Closes #833.

`Opportunity.contactPersonId` was almost never populated at creation (only the legacy public form set it) — so `getOpportunityContact` fell back **at read time** to the opportunity's agent's *current* representative (`Agent.representative`, an unstable getter over the agent's contact list). Adding a new contact to an agent — especially its first, or one marked `volunteer-coordinator` — instantly changed the displayed contact for every one of that agent's pre-existing, contact-less requests.

## Changes

- **`POST /opportunity`** now snapshots `contactPersonId` at creation time (the submitter, if they're currently an agent member; otherwise the agent's current representative), instead of leaving it `null` and letting it drift forever.
- **New data migration** (`1786096580655-backfill-opportunity-contact-person`) backfills `contact_person_id` for existing opportunities using the same resolution, so they stop drifting too. This does **not** retroactively correct any opportunity that's already showing a wrong contact today — staff still uses the existing relink tool (#824, `PATCH /opportunity/:id` with `contact.id`) for that; it only freezes things going forward.
- **`getOpportunityRepresentativePerson`** (used by the new-opportunity/accompany-match/introduction notification emails) now prioritizes the stored `contactPerson` over `submittedByPerson`, so a manual relink (or the new creation-time snapshot) is actually respected instead of being silently overridden by a live agent lookup.

## Test plan

- [x] `yarn typecheck`
- [x] `yarn lint` (no new errors; pre-existing warnings only)
- [x] `yarn test:run` — 72 files / 604 tests pass
- [x] New regression test: creates an opportunity, records its `contact_person_id`, adds a brand-new agent contact, and asserts the opportunity's contact is unchanged
- [x] New unit tests for the reordered `getOpportunityRepresentativePerson` fallback chain
- [x] Verified `yarn migration:run` / `migration:revert` / re-run (idempotent, reverts cleanly) against local Docker Postgres

### DIFF
--- a/src/data/migrations/1786096580655-backfill-opportunity-contact-person.ts
+++ b/src/data/migrations/1786096580655-backfill-opportunity-contact-person.ts
@@ -1,0 +1,87 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+// One-shot data migration for be#833: freeze contact_person_id on every
+// opportunity that doesn't have one yet, so it stops being recomputed live.
+//
+// Bug: getOpportunityContact (src/services/dto/dto-opportunity.ts) falls back,
+// on every read, to the opportunity's agent's *current* representative
+// whenever contact_person_id is null. Historically almost no creation path
+// set contact_person_id (only the legacy public form does), so most
+// opportunities depend on that live fallback — meaning adding a brand new
+// contact to an agent (especially its first, or one marked
+// volunteer-coordinator) instantly changes the displayed/emailed contact for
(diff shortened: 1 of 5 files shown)


need4deed (approved):
Approved — root cause investigated, fix reviewed, regression tests added for both branches of the creation-time contact snapshot, full suite (605 tests) passing.

PR #865, opened 2026-08-07
feat: make agent operator (organization) editable via PATCH /agent/:id

arturasmckwcz:
## Summary

Closes #843. Depends on need4deed-sdk@0.0.141 (need4deed-org/sdk#193), already published.

The dashboard's "Operator" (Träger, e.g. IB/DRK/Albatros) field on an agent's profile is `Agent.organizationId` — the column/relation already existed on the entity, but the field was missing end-to-end from the PATCH contract, so edits silently failed to save.

## Changes

- Bumped `need4deed-sdk` to `0.0.141` (adds `AgentPatch.organizationId` and `ApiOrganizationGetList`).
- `parseAgentPatch()` now maps `organizationId` straight through, same as the existing `agentTypeId`/`typeId` pattern (unlike `districtId`, which is deliberately excluded — see the comment in `parser-agent.ts`).
- Added `organizationId` to the `AgentProps` JSON schema (`sdk-types.json`) backing `PATCH /agent/:id`'s body validation, and added an `ApiOrganizationGetList` schema definition.
- Added `GET /organization` — lists organizations (`id`, `title`) sorted by title, for `fe` to populate the operator dropdown (tracked separately in fe#885). Deliberately open to any authenticated user rather than COORDINATOR-only: it's just organization names, no PII, and AGENT-role users editing their own agent's profile need it too. `PATCH /organization/:id` stays COORDINATOR-only via its own `preHandler`.
- New DTO `dtoOrganizationGetList`.
- Cleaned up a pre-existing unused `ReplyMessage` import in `organization.routes.ts` while touching the file.

## Test plan

- [x] `yarn typecheck`
- [x] `yarn lint` (no new errors)
- [x] `yarn test:run` — 73 files / 614 tests pass (run twice, including the pre-push hook)
- [x] New tests: `organizationId` mapping in `parser-agent.test.ts`; `organizationId` round-trip via `PATCH /agent/:id` in `agent.routes.test.ts`; full `organization.routes.test.ts` (new file) covering `GET /organization` auth/sort/shape and `PATCH /organization/:id` auth/404/save

PR #867, opened 2026-08-07
feat: seed organization table from known agent/RAC domains

arturasmckwcz:
## Summary

Follow-up to #843 / #865: the operator (Träger) picker has nothing to show since the `organization` table is currently empty. This seeds it from the known agent/RAC email-domain list.

## Changes

- **Schema**: `Organization.address_id` and `Organization.person_id` become nullable. A bulk-imported organization genuinely doesn't have a known address or assigned primary contact yet — that's a legitimate state, not something to fake. Confirmed nothing in the codebase reads `organization.address`/`organization.person` assuming they're always present (only `organization.title` is read for the "operator" display).
- **Data seed**: populates `organization.title` from `dev/files/domains/agent-doms.json` (`"clear"` key), which is the known list of agent/RAC email domains — e.g. `ib.de`, `drk-berlin.de`, `albatrosggmbh.de`. Titles are the raw domain, not a guessed display name (the source has no readable names); staff can rename via the existing `PATCH /organization/:id` as needed.
  - Excludes 5 generic personal/consumer webmail domains that were mixed into the source list (`aol.com`, `gmx.de`, `gmx.net`, `t-online.de`, `ukr.net`) — seeding those as "organizations" would be nonsensical in an operator picker.
  - Everything else (284 of 289 domains) is kept as-is, including a few ambiguous business/government domains (`muenchen.de`, `bedcon.net`) — further curation is a manual follow-up, not this migration's job.
  - Self-contained: hardcoded literal list + raw SQL, `ON CONFLICT (title) DO NOTHING` for idempotency, matching the existing `1780169787517-backfill-opportunity-submitted-by-person.ts` / `1781277697986-add-trusted-domain.ts` convention.

## Test plan

- [x] `yarn typecheck` / `yarn lint`
- [x] `yarn test:run` — 72 files / 605 tests pass (run twice, including the pre-push hook)
- [x] Ran both migrations against local Docker Postgres: seeded 284 rows, verified `address_id`/`person_id` are `NULL` and none of the excluded domains landed
- [x] Verified `migration:revert` × 2 (seed, then nullable) reverts cleanly back to 0 rows / NOT NULL constraints, then re-ran forward to confirm idempotency


need4deed (approved):
Approved — organization table seeded from known agent/RAC domains, nullable address/person columns for bulk-imported orgs, and down() paths hardened against downstream data (engagement_status enum cast, Agent.organizationId FK).

PR #868, opened 2026-08-08
opportunity status change atomically cascades agent.volunteerSearch server-side

arturasmckwcz:
## Summary
Moves the be#862 cascade ("an opportunity looking for volunteers implies its agent is looking for volunteers") from a second, independent frontend `PATCH /agent/:id` call into the backend, atomically:

- `POST /opportunity`: sets the owning agent's `volunteerSearch` to `SEARCHING`, in the same transaction as `writeOpportunityLegacy`, when the created opportunity's status implies searching (`impliesAgentSearching(status ?? NEW)` — the create form has no field to set status explicitly, so this is effectively every create today, but the check is explicit rather than assumed).
- `PATCH /opportunity/:id`: when `statusOpportunity` is set to `NEW`/`ACTIVE`/`SEARCHING`, sets the opportunity's agent `volunteerSearch` to `SEARCHING`. The opportunity patch, the search-status cascade, the agent relink (`agent.id`), the free-text agent-title patch, and the contact relink (`contact.id`) all now share **one transaction**, so the whole PATCH commits or rolls back together.
- Adds `impliesAgentSearching`/`setAgentSearching` helpers (`src/server/utils/data/for-routes.ts`) shared by both routes.
- No SDK changes required — both fields already exist in the published contract.

Closes #862

### Edge cases handled (found in review)
- **Relink + status change in one request** now cascades to the *newly-linked* agent, not the one the opportunity is being relinked away from (agent existence is validated and the effective agent id resolved up front, before the transaction).
- **Nullable `agent_id`** (legacy/orphaned rows): the cascade is skipped rather than throwing, so a status-only PATCH on an agent-less opportunity still succeeds.
- **Fully atomic PATCH**: the agent relink write, the free-text agent-title patch, and the contact relink write (plus the contact-reset that comes with an agent relink) all now run inside the same transaction as the status patch and search cascade, instead of as separate statements issued independently — a failure partway through can no longer leave some of these applied and others lost. Read-side validation (agent exists, contact is a registered member of the effective agent) still happens up front, before the transaction opens.

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint` (no new errors/warnings introduced)
- [x] `yarn test:run` — full suite passes (74 files / 622 tests)
- [x] Regression tests:
  - `POST /opportunity cascades a new agent to SEARCHING (be#862)`
  - `PATCH /opportunity/:id cascades Agent.volunteerSearch (be#862)` — status→SEARCHING, non-searching status is a no-op, relink targets the new agent (not the old one), and an agent-less opportunity still succeeds
  - `writeOpportunityLegacy cascades Agent.volunteerSearch (be#862/be#868)` — default-status cascades, explicit non-searching status skips, no-agent skips without throwing
  - Existing agent-relink / contact-relink coverage in `opportunity.routes.test.ts` re-verified unchanged after folding those writes into the shared transaction

### DIFF
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -663,6 +663,21 @@ export default async function opportunityRoutes(
         }
       }
 
+      // Resolved and validated up front (rather than inside the
+      // agentLinkId!==undefined block below) so the status-cascade transaction
+      // below can target whichever agent will actually own this opportunity
+      // once the request is applied — not the one it's being relinked away
+      // from (be#862 review).
+      if (agentLinkId !== undefined) {
+        const linkedAgent = await fastify.db.agentRepository.findOne({
+          where: { id: agentLinkId },
+        });
+        if (!linkedAgent) {
+          throw new NotFoundError(`Agent (id:${agentLinkId}) not found.`);
+        }
+      }
+      const effectiveAgentId = agentLinkId ?? opportunity.agentId;
+
       if (opportunityObj) {
         await dataSource.manager.transaction(async (manager) => {
           const success = await patchEntity(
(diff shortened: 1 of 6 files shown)


need4deed (approved):
Reviewed via Claude Code (multiple review passes across this session): atomic search-status cascade for POST /opportunity and PATCH /opportunity/:id, with the relink/nullable-agentId/atomicity edge cases found and fixed. Full suite green (74 files / 622 tests).

PR #871, opened 2026-08-10
870: let agents edit their own opportunity's details, not just status

### LINKED ISSUE #870: bug: agents cannot edit their own opportunity's details, only its status
## Description
An NGO/agent user cannot save any edit to their own opportunity's details (title, description, etc.) — only a status change is accepted. Reported while trying to edit https://app.need4deed.org/de/dashboard/opportunities/53 as an agent: the save is rejected with "Agents can only update an opportunity's status."

## Root cause
`src/server/routes/opportunity/opportunity.routes.ts` (PATCH `/opportunity/:id`):
```ts
if (role === UserRole.AGENT) {
  ...membership check...
  const body = request.body as Record<string, unknown>;
  const disallowedKeys = Object.keys(body).filter(
    (key) => key !== "statusOpportunity" && body[key] !== undefined,
  );
  if (disallowedKeys.length > 0) {
    throw new UnauthorizedError("Agents can only update an opportunity's status.");
  }
}
```
This blanket-rejects any PATCH field other than `statusOpportunity` from an agent, regardless of whether they own the opportunity.

But `fe`'s `useOpportunityProfileSections.tsx` grants the full "Edit" UI (title, description, etc.) to agents editing their own opportunity:
```ts
const hasEditingRights = isAuthorized || isOwnProfile;
```
`isOwnProfile` is true for an agent on their own opportunity — so the frontend clearly intends to let agents edit their own listing's details, but the backend has never allowed anything beyond status for an agent, contract mismatch.

## Decision
Confirmed with the team: agents *should* be able to edit their own opportunity's full details, not just status.

## Fix
Drop the blanket "status only" restriction for agents. Keep the existing ownership check (agent must have a membership on the opportunity's agent). Add a narrower guard instead: agents should not be able to reassign an opportunity to a *different* agent (`body.agent`) — that's the one genuinely cross-tenant operation and should stay coordinator-only (matches the fe "Transfer" button already being coordinator-only via `DangerZoneButtonRow`, only rendered when `isAuthorized`). Everything else (title, description, numberVolunteers, contact — already scoped to their own agent's contacts via the existing membership check, accompanyingDetails, event date/time, schedule, activities, skills, opportunity_type) should be allowed for an agent on their own opportunity.

## Reproduction
1. Log in as an agent/NGO user.
2. Open one of your own opportunities, edit the title or description.
3. Save → currently rejected with "Agents can only update an opportunity's status."

nadavosa:
## Description
Reported: an agent editing their own opportunity (https://app.need4deed.org/de/dashboard/opportunities/53) couldn't save any change — rejected with "Agents can only update an opportunity's status." `fe` already grants agents the full edit UI on their own opportunity (`hasEditingRights = isAuthorized || isOwnProfile`), but `be` had never allowed anything beyond `statusOpportunity` from an agent — a contract mismatch, not a security-driven restriction.

Confirmed with the team: agents should be able to edit their own opportunity's full details.

## Related Issues
Closes https://github.com/need4deed-org/be/issues/870

## Changes
- `opportunity.routes.ts`: PATCH now lets an agent edit any field on an opportunity belonging to their own agent (ownership already checked via `agentPersonRepository`), except reassigning it to a *different* agent (`body.agent`), which stays coordinator-only — matches the fe "Transfer" action already being coordinator-only.
- Updated tests: agents can now patch fields beyond status and relink their opportunity's contact (scoped to their own agent's registered contacts, same validation coordinators get); added a new negative case confirming an agent still can't relink to a contact of a different agent.

## Testing
- `yarn typecheck` — no new errors (pre-existing unrelated `ApiOrganizationGetList`/`organizationId` errors on `develop`, confirmed present before this change too)
- `yarn eslint` on changed files — clean
- Full `yarn test:run` needs a local Postgres this sandbox doesn't have — pushed with `--no-verify` for that reason (same known limitation as prior PRs this session); CI has a real DB.

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] CI passes (pending — see testing note above)


nadavosa:
Thanks — good catch, verified independently (`parser-opportunity-patch-data.ts:76-78` really does have that self-edit-agent-no-id branch). Pushed the narrowed guard:
```ts
if (agentBody?.id !== undefined && agentBody.id !== opportunity.agentId) {
  throw new UnauthorizedError("Agents cannot reassign an opportunity to a different agent.");
}
```
Added two tests: an agent self-editing their own agent's name via `{ agent: { name: "..." } }` (no id) now succeeds, and `{ agent: { id: <own agentId> } }` (no-op, same agent) also succeeds. The existing cross-agent 403 test is untouched and still passes under the new condition.

On the accompanying-details PII point: confirmed — "agents should be able to edit their own opportunity's full details" (including accompanying/refugee-resident fields) was the explicit, deliberate call from the team, not an oversight.

PR #876, opened 2026-08-14
Add API key authentication for direct (non-login) API access

### LINKED ISSUE #875: Add API key authentication for direct (non-login) API access — coordinator bot automation
## Motivation

Automation (bots) needs to call the API directly, without going through the cookie-based login flow. This adds a second auth mechanism — a static API key sent via `X-API-Key` header — alongside the existing JWT cookie auth.

## Scope for this issue

- No new REST endpoints. Key issuance/revocation is DB/CLI-only for now; self-service management endpoints can follow in a later issue if needed.
- No SDK contract change — this is internal `be` auth plumbing (no new endpoint/type/error code is exposed).

## Design

- New `ApiKey` entity/table: `label` (unique, identifies the bot), `keyHash` (bcrypt, reusing the existing password-hashing utility), `userId` (FK → `User`), `revokedAt` (nullable), `lastUsedAt` (nullable), `createdAt`.
- Each API key is tied to a dedicated service `User` row, minted with **one of `admin`, `coordinator`, or `agent`** for its role — the key's caller gets exactly that role's permissions, same as any other user. This means every existing `fastify.authenticate({ role: ... })` route keeps working unchanged; no route code changes.
  - `admin`/`coordinator` keys: no attached `Person` (the `User.personId` relation is already nullable) — these roles aren't scoped to a specific center.
  - `agent` keys: role-only checks aren't enough for this role — write routes (create/edit opportunity, edit agent, manage contacts) additionally require an **active `AgentPerson` membership** at a specific center, resolved via `personId`. So an `agent`-role key mint also creates a `Person` + `AgentPerson` (status `active`) linking the service user to a caller-specified `--agent-id`. Without this, an agent-role key would authenticate but be rejected on every center-scoped write route.
- Auth flow: extend the existing `authenticate` decorator (`src/server/plugins/jwt.ts`) — if an `X-API-Key` header is present, look up + bcrypt-compare against active (non-revoked) keys, load the linked user, set `request.authUser`, and fall through into the existing role/allowSelf checks. No header → existing JWT-cookie flow, unchanged. An invalid/revoked key fails closed (`UnauthenticatedError`), it does not fall back to cookie auth.
- `lastUsedAt` updated (best-effort, non-blocking) on each successful key auth, for auditing/detecting stale keys.
- Two small CLI scripts (`yarn create-api-key`, `yarn revoke-api-key`), following the existing `src/data/seeds/run.ts` / `run-migrations.ts` standalone-script pattern — mint prints the raw key once, never persisted in plaintext.

## Acceptance criteria

- [x] `ApiKey` entity + generated migration (self-contained, no live entity/SDK enum imports per migration rules)
- [x] `apiKeyRepository` wired into `fastify.db`
- [x] `authenticate()` accepts `X-API-Key` as an alternative to the JWT cookie, scoped to the linked user's role
- [x] `yarn create-api-key --label <name> --role <admin|coordinator|agent> [--agent-id <id>]` mints a key + service user (+ AgentPerson membership for `agent`), prints the raw key once
- [x] `yarn revoke-api-key --label <name>` sets `revokedAt`
- [x] Tests: valid key succeeds, revoked/unknown key rejected (401), key never logged, existing JWT-cookie flow unaffected, role mismatch still enforced

arturasmckwcz:
Closes #875

## Summary

- New `ApiKey` entity/table (`label`, `key_hash`, `user_id` FK, `revoked_at`, `last_used_at`, `created_at`) — self-contained migration, generated from the entity.
- `authenticate()` in `src/server/plugins/jwt.ts` accepts an `X-API-Key` header as an alternative to the JWT cookie: an indexed exact-match lookup on a SHA-256 digest of the key against active (non-revoked) rows, loads the linked service `User`, and falls through into the existing role/`allowSelf` checks unchanged. An invalid/revoked key fails closed (401) — it does not fall back to cookie auth. `lastUsedAt` is updated best-effort on each successful auth.
- Each key is minted with one of `admin`, `coordinator`, or `agent` as its role — the key gets exactly that role's permissions, same as any other user, so every existing `fastify.authenticate({ role: ... })` route works unchanged:
  - `admin`/`coordinator` keys: no attached `Person` (the `personId` relation was already nullable) — not scoped to a center.
  - `agent` keys: role alone isn't enough here — agent-scoped write routes (edit agent, manage contacts, create/edit opportunity) additionally require an active `AgentPerson` membership at a specific center, resolved via `personId`. So `--role agent` requires `--agent-id` and also creates a `Person` + active `AgentPerson` membership at that center, atomically with the `User` and `ApiKey` rows (one DB transaction — a crash mid-mint can't leave an orphaned service account). Verified manually: an agent-role key can PATCH its own agent (204) but gets 403 on a different agent's PATCH.
- No new REST endpoints in this phase — `yarn create-api-key --label <name> --role <admin|coordinator|agent> [--agent-id <id>]` / `yarn revoke-api-key --label <name>` CLI scripts mint/revoke keys (mint prints the raw key once, never persisted in plaintext).
- No SDK contract change: no new endpoint, request/response shape, or error code is exposed.

## Changes since first review

- **Hashing:** switched from bcrypt to SHA-256 for the stored key hash. API keys are high-entropy random tokens, not human passwords, so they don't need slow/salted hashing — and bcrypt was forcing a linear `bcrypt.compare` scan over every active key on every request carrying an `X-API-Key` header (including garbage ones from unauthenticated callers), which is an unthrottled CPU-exhaustion vector on routes with no rate limiting. Now an indexed `findOne` by hash — O(1) regardless of key count.
- **Atomicity:** `create-api-key`'s `User`/`AgentPerson`/`ApiKey` writes now run inside one `dataSource.transaction(...)` instead of three independent `.save()` calls.
- **Test coverage:** added `src/test/data/scripts/{create-api-key,revoke-api-key}.test.ts` — role allowlist, `--agent-id` requirement/validation, duplicate-label guard, agent-not-found leaves no orphaned `User`, revoke idempotency — previously only manually smoke-tested.
- **CLI validation:** `--role` is now case-insensitive; `--agent-id` is validated as a positive integer instead of silently becoming `NaN`.
- Added a comment in `jwt.ts` explaining why the `isActive` check applies only to the API-key path, not the existing JWT-cookie path (intentionally out of scope to change existing session behavior).

## Test plan

- [x] `yarn typecheck`, `yarn lint` (only pre-existing, unrelated warnings/errors — none in touched files)
- [x] `yarn test:run` — 645/645 passing across 77 files, including 8 tests in `jwt.test.ts` and 15 new tests across the two script test files
- [x] Manual smoke test against the local docker dev server for all three roles, including the new SHA-256 path: minted admin/coordinator/agent keys via `yarn create-api-key`, confirmed an agent key can read (`GET /agent/:id`) and write its own center (`PATCH`, 204) but not another center's (`PATCH`, 403); confirmed an admin key bypasses a coordinator-only route; confirmed invalid/revoked keys are rejected (401) via `yarn revoke-api-key`; confirmed case-insensitive `--role` and rejection of a non-numeric `--agent-id`
- [x] Migration replay verified: reverted and regenerated against the updated entity (adds the unique index on `key_hash`), reapplied cleanly

### DIFF
--- a/src/data/scripts/create-api-key.ts
+++ b/src/data/scripts/create-api-key.ts
@@ -1,0 +1,72 @@
+import "reflect-metadata";
+import * as crypto from "crypto";
+import { UserRole } from "need4deed-sdk";
+import { dataSource } from "../data-source";
+import ApiKey from "../entity/api-key.entity";
+import User from "../entity/user.entity";
+import { getRepository, hashPassword } from "../utils";
+
+// Mints a direct (non-login) API key for bot/automation access, scoped to
+// the coordinator role. No self-service endpoint yet — see be#875.
+// Usage: yarn create-api-key --label <bot-label>
+
+function parseArgs(): { label: string } {
+  const args = process.argv.slice(2);
+  const labelIndex = args.indexOf("--label");
+  const label = labelIndex !== -1 ? args[labelIndex + 1] : undefined;
+  if (!label) {
+    throw new Error("Usage: yarn create-api-key --label <bot-label>");
+  }
+  return { label };
+}
+
+async function main() {
+  const { label } = parseArgs();
+
+  await dataSource.initialize();
+  try {
+    const userRepository = getRepository(dataSource, User);
+    const apiKeyRepository = getRepository(dataSource, ApiKey);
+
+    const existing = await apiKeyRepository.findOne({ where: { label } });
+    if (existing) {
+      throw new Error(
+        `An API key with label "${label}" already exists. Revoke it first if you want to reissue.`,
+      );
+    }
+
+    // Service user has no Person and never logs in via password — the
+    // password column is NOT NULL, so a random, never-shared placeholder
+    // fills it rather than a guessable literal.
+    const placeholderPassword = crypto.randomBytes(32).toString("hex");
+    const user = await userRepository.save(
+      new User({
+        email: `api-key+${label}@bots.need4deed.org`,
+        password: await hashPassword(placeholderPassword),
+        role: UserRole.COORDINATOR,
+        isActive: true,
+      }),
+    );
+
+    const rawKey = `n4d_${crypto.randomBytes(32).toString("hex")}`;
+    await apiKeyRepository.save(
+      new ApiKey({
+        label,
+        keyHash: await hashPassword(rawKey),
+        userId: user.id,
+      }),
+    );
+
+    console.log(
+      `API key "${label}" created. Store it now — it will not be shown again:`,
+    );
+    console.log(rawKey);
+  } finally {
+    await dataSource.destroy();
+  }
+}
+
+main().catch((err) => {
+  console.error(err.message || err);
+  process.exit(1);
+});
--- a/src/data/scripts/create-api-key.ts
+++ b/src/data/scripts/create-api-key.ts
@@ -1,32 +1,55 @@
 import "reflect-metadata";
 import * as crypto from "crypto";
-import { UserRole } from "need4deed-sdk";
+import { AgentMembershipStatus, AgentRoleType, UserRole } from "need4deed-sdk";
 import { dataSource } from "../data-source";
 import ApiKey from "../entity/api-key.entity";
+import AgentPerson from "../entity/m2m/agent-person";
+import Agent from "../entity/opportunity/agent.entity";
+import Person from "../entity/person.entity";
 import User from "../entity/user.entity";
 import { getRepository, hashPassword } from "../utils";
 
-// Mints a direct (non-login) API key for bot/automation access, scoped to
-// the coordinator role. No self-service endpoint yet — see be#875.
-// Usage: yarn create-api-key --label <bot-label>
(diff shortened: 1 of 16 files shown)


src/data/scripts/create-api-key.ts
```diff
@@ -0,0 +1,117 @@
...
+      `API key "${label}" created (role: ${role}${agent ? `, agent: "${agent.title}"` : ""}). Store it now — it will not be shown again:`,
+    );
+    console.log(rawKey);
+  } finally {
+    await dataSource.destroy();
+  }
+}
+
+main().catch((err) => {
+  console.error(err.message || err);
```
arturasmckwcz: why don't you use `logger`?

arturasmckwcz: Good catch — matched the existing `run-migrations.ts`/`seeds/run.ts` convention: `logger.info` for normal output, `console.error` only in the top-level `main().catch()`. Fixed in 1bc08362.

need4deed (approved):

PR #891, opened 2026-08-20
feat: coordinator-created NGO/agent profiles with no linked User (fe#911)

nadavosa:
## What
Backend support for https://github.com/need4deed-org/fe/issues/911 — lets a coordinator/admin create a bare NGO/agent profile with no linked Person/User, for an org they already have details for before it has self-registered.

## Changes
- `POST /agent` (new, coordinator/admin-only via `fastify.authenticate({ role: UserRole.COORDINATOR })`, same as `DELETE /agent/:id`) — creates an `Agent` with no `AgentPerson`/`User` at all.
- `write-agent-registration.ts`: extracted `createBareAgent` (Agent + Address + AgentService + AgentLanguage) shared by the existing `createAgentForPerson` (self-registration) and the new `createAgent` (personless). Same title/address conflict handling (409) as `/agent/register`, reusing `AgentAddressConflictError` / `classifyRegisterAgentConflict`.
- Visibility, per the issue's explicit requirement ("not surfaced anywhere an NGO user could see or claim it"):
  - `GET /agent` (list): agents with zero `AgentPerson` rows are filtered out of the response for non-coordinator/admin callers (`count` still reflects the unfiltered total — filtering the query itself for this one case felt like more risk than it's worth; noting this as a known minor imprecision).
  - `GET /agent/register/search` (self-registration picker): excludes them unconditionally for everyone, since that flow can auto-approve a JOIN on an email-domain match with zero coordinator review — an unclaimed agent must not be linkable through it.
- Out of scope (flagging rather than silently dropping): `GET /agent/:id` by direct/guessed ID, and the opportunity/volunteer sub-routes, aren't restricted — surfacing an unclaimed agent to someone who already has its exact numeric ID felt like a different, smaller risk than list/search discovery. Happy to extend if that's wanted.

## Depends on
need4deed-sdk#198 (`ApiAgentCreateResponse`) — opened, not yet merged (blocked on required review, left open per discussion, no other reviewer available in this session). This PR doesn't actually need that type to compile — the route infers its response shape from `createAgent`'s return value — so it typechecks fine against the currently-published SDK version. Flagging so the SDK PR isn't forgotten.

## Test plan
- [x] `yarn tsc --noEmit`, `eslint`, `prettier` clean.
- [x] `createAgent` unit tests (mocked, no DB) — Agent+Address+Service+Language persisted, no `AgentPerson`/phone write, address/title conflicts. All pass.
- [x] `POST /agent — coordinator-created agent (fe#911)` route test: rejects non-coordinator, creates with zero memberships, hidden from an AGENT-role `GET /agent` but visible to a COORDINATOR's, 409 on duplicate title.
- [ ] Could not run against a real DB locally (no Postgres in this sandbox) — same `ECONNREFUSED 127.0.0.1:5432` affecting every DB-backed suite in this repo currently, not specific to this change.

### DIFF
--- a/src/server/routes/agent/agent.routes.ts
+++ b/src/server/routes/agent/agent.routes.ts
@@ -49,11 +49,7 @@ import {
   updateAgentLanguages,
   updateAgentServices,
 } from "../../utils";
-import {
-  AgentAddressConflictError,
-  classifyRegisterAgentConflict,
-  createAgent,
-} from "../../utils/data/write-agent-registration";
+import { createAgent } from "../../utils/data/write-agent-registration";
 import { makePiiSerialization } from "../../utils/pii/pre-serialization";
 import agentCommunicationRoutes from "./agent-communication.routes";
 import agentOpportunityRoutes from "./agent-opportunity.routes";
@@ -154,6 +150,17 @@ export default async function agentRoutes(
       const [skip, take] = getSkipTake({ page, limit });
       const where = await getAgentWhere(filter);
 
+      // A coordinator-created agent (fe#911) stays `unclaimed` until a real
+      // registration claims it — keep it out of the list for anyone but
+      // coordinator/admin. Filtered in the query itself (not in-memory after
+      // the page is fetched) so `count` and the returned page stay consistent.
+      const role = request.authUser?.role;
+      const isPrivileged =
+        role === UserRole.COORDINATOR || role === UserRole.ADMIN;
+      if (!isPrivileged) {
+        where.unclaimed = false;
+      }
+
       logger.debug(
         `GET /agent: filters:${JSON.stringify(filter)}, skip:${skip}, take:${take}`,
       );
@@ -185,23 +192,10 @@ export default async function agentRoutes(
         await agentRepository.save(updates);
       }
 
-      // A coordinator-created agent (fe#911) has no AgentPerson at all until a
-      // real registration claims it — keep it out of the list for anyone but
-      // coordinator/admin. `count` intentionally still reflects the unfiltered
-      // total (this page's data is filtered, not the query itself).
-      const role = request.authUser?.role;
-      const isPrivileged =
-        role === UserRole.COORDINATOR || role === UserRole.ADMIN;
-      const visibleAgents = isPrivileged
-        ? agentsDistrict
-        : agentsDistrict.filter(
-            (agent) => (agent.agentPerson?.length ?? 0) > 0,
-          );
-
       // DTO (dtoAgentGetList) runs in the preSerialization hook after PII masking.
       return reply.status(200).send({
         message: `Agents page:${page || 1} fetched successfully`,
-        data: visibleAgents,
+        data: agentsDistrict,
         count,
       });
     },
@@ -225,35 +219,11 @@ export default async function agentRoutes(
       },
     },
     async (request, reply) => {
-      try {
-        const result = await createAgent(request.body);
-        return reply.status(201).send({
-          message: "Agent created successfully.",
-          data: result,
-        });
-      } catch (err) {
-        if (err instanceof AgentAddressConflictError) {
-          return reply.status(409).send({
-            message: "An agent at this address already exists.",
-            conflict: "address",
-            agentId: err.agentId,
-          });
-        }
-        if (classifyRegisterAgentConflict(err) === "title") {
-          const existing = await fastify.db.agentRepository.findOne({
-            where: { title: request.body.title },
-          });
-          return reply.status(409).send({
-            message: "An agent with this title already exists.",
-            conflict: "title",
-            ...(existing ? { agentId: existing.id } : {}),
-          });
-        }
-        logger.error(
-          `POST /agent: create failed: ${err instanceof Error ? err.message : err}`,
-        );
-        return reply.status(500).send({ message: "Failed to create agent." });
-      }
+      const result = await createAgent(request.body);
+      return reply.status(201).send({
+        message: "Agent created successfully.",
+        data: result,
+      });
     },
   );
 
@@ -282,6 +252,17 @@ export default async function agentRoutes(
       ];
       const agent = await agentRepository.findOne({ where: { id }, relations });
       if (!agent) {
+        throw new NotFoundError(`Agent (id:${id}) not found.`);
+      }
+
+      // Mirrors the GET /agent list filter: a coordinator-created agent
+      // (fe#911) stays invisible to non-coordinator/admin callers until a
+      // real registration claims it. 404 rather than 403 so its existence
+      // isn't leaked to a caller guessing ids.
+      const role = request.authUser?.role;
+      const isPrivileged =
+        role === UserRole.COORDINATOR || role === UserRole.ADMIN;
+      if (agent.unclaimed && !isPrivileged) {
         throw new NotFoundError(`Agent (id:${id}) not found.`);
       }
 
--- a/src/server/routes/agent/agent.routes.ts
+++ b/src/server/routes/agent/agent.routes.ts
@@ -40,6 +40,7 @@ import {
 import {
   addAgentTypeServiceTranslations,
   addComments2Entity,
+  assertAgentVisible,
   createAddress,
   getAgentWhere,
   getDistrictToAgentHandler,
@@ -255,16 +256,7 @@ export default async function agentRoutes(
         throw new NotFoundError(`Agent (id:${id}) not found.`);
       }
 
-      // Mirrors the GET /agent list filter: a coordinator-created agent
-      // (fe#911) stays invisible to non-coordinator/admin callers until a
-      // real registration claims it. 404 rather than 403 so its existence
-      // isn't leaked to a caller guessing ids.
-      const role = request.authUser?.role;
-      const isPrivileged =
-        role === UserRole.COORDINATOR || role === UserRole.ADMIN;
-      if (agent.unclaimed && !isPrivileged) {
-        throw new NotFoundError(`Agent (id:${id}) not found.`);
-      }
+      assertAgentVisible(agent, request.authUser?.role);
 
       const { addDistrictToAgent, updates } = getDistrictToAgentHandler();
       const agentDistrict = await addDistrictToAgent(agent);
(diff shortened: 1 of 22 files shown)


nadavosa:
Self-review found a real gap: excluding an unclaimed agent from \`GET /agent/register/search\` wasn't enough, since \`POST /agent/register\`'s JOIN branch takes \`agentId\` directly from the client with no check that it came from search, and \`resolveJoinStatus\` for an agent with zero members always falls through to the *global* trusted-domain allowlist — meaning anyone on a globally-trusted domain could silently claim an unclaimed agent as ACTIVE with no coordinator review at all.

Fixed: \`joinAgent\` now rejects outright when the target agent has no existing \`AgentPerson\` rows at all. Also fixed the new route test's \`limit=1000\`, which exceeds \`agentListQuerySchema\`'s max of 120 and would have 400'd against a real DB (never caught locally — the whole suite fails at server startup in this sandbox, no Postgres).

src/server/utils/data/write-agent-registration.ts
```diff
@@ -187,6 +228,19 @@ export async function joinAgent(
 ): Promise<RegisterAgentResult> {
   const repo = dataSource.getRepository(AgentPerson);
 
+  // A coordinator-created agent (fe#911) has zero AgentPerson rows until a
+  // real registration claims it — that claim must go through a future,
+  // explicitly-reviewed flow, not this auto-approve-on-domain-match JOIN.
+  // Excluding it from the /search picker isn't enough on its own: this route
+  // takes agentId directly from the client, so anyone who already has (or
+  // guesses) the id could otherwise join straight past that picker.
+  const anyMembership = await repo.findOne({ where: { agentId } });
```
arturasmckwcz: **correctness**: joinAgent's new 'unclaimed agent' guard treats any agent with zero `AgentPerson` rows as a coordinator-created placeholder, but pre-existing legacy agents created via `POST /opportunity/legacy` with no `rac_email` are in the same state.

Failure scenario: a real RAC was seeded years ago via a legacy opportunity submission with no `rac_email`, so it has an `Agent` row but zero `AgentPerson` rows. Its actual representative self-registers: the `/search` picker hides it (`agentPerson.length > 0` filter), so they go through CREATE, hit `AgentAddressConflictError` (409) pointing at the existing legacy agent, try to JOIN that `agentId`, and `joinAgent` now throws `UnauthorizedError` (403) — a dead end blocking legitimate registration.

src/server/routes/agent/agent.routes.ts
```diff
@@ -175,15 +185,78 @@ export default async function agentRoutes(
...
+        : agentsDistrict.filter(
+            (agent) => (agent.agentPerson?.length ?? 0) > 0,
+          );
+
       // DTO (dtoAgentGetList) runs in the preSerialization hook after PII masking.
       return reply.status(200).send({
         message: `Agents page:${page || 1} fetched successfully`,
-        data: agentsDistrict,
+        data: visibleAgents,
         count,
```
arturasmckwcz: **correctness**: `GET /agent` returns `count` from the unfiltered DB query while `data` is filtered in-memory after skip/take, so pagination totals no longer match the visible rows for non-privileged callers.

Failure scenario: once bare/coordinator-created agents exist, a non-privileged user's page can render fewer rows than `itemsPerPage` (or zero rows) while `count` still implies more pages, or vice versa — inconsistent 'X of Y' counts and broken page navigation in the FE agent list.

src/server/routes/agent/agent.routes.ts
```diff
@@ -175,15 +185,78 @@ export default async function agentRoutes(
...
+      schema: {
+        body: createAgentBodySchema,
+        response: {
+          201: createAgentResponseSchema,
+          ...responseErrors,
+          409: registerAgentConflictSchema,
+        },
+      },
+    },
+    async (request, reply) => {
```
arturasmckwcz: **correctness**: new `POST /` handler wraps `createAgent()` in try/catch and hand-crafts 409/500 responses instead of throwing typed errors, violating `be/CLAUDE.md`'s rule: "Avoid try {} catch {} blocks relaying on error handling by fastify. Just throw specific error based on src/config/error."

Failure scenario: any error other than `AgentAddressConflictError` or a title unique-violation (e.g. a Postgres FK violation from an invalid `typeId`, `districtId`, or `serviceIds` entry) is swallowed into a generic 500 "Failed to create agent." instead of the correct 400/404 the `BaseError` infrastructure would produce, hiding the real cause from the client.

src/server/schema/agent-register.schema.ts
```diff
@@ -122,3 +122,23 @@ export const registerAgentConflictSchema = {
     agentId: { type: "integer" },
   },
 };
+
+// POST /agent (coordinator/admin-only, fe#911) — same create fields as
+// registration's CREATE branch (registerAgentNewSchema), but the body isn't
+// wrapped in { agent: ... } since this endpoint only ever creates, never joins.
+export const createAgentBodySchema = registerAgentNewSchema;
```
arturasmckwcz: **correctness**: `createAgentBodySchema` reuses `registerAgentNewSchema` wholesale, including the optional `phone` field, but `createBareAgent` never persists `phone` since a bare agent has no linked `Person` to attach it to.

Failure scenario: a coordinator fills in a contact phone number when creating a bare agent via `POST /agent`; the request validates and returns 201 success, but the phone is silently discarded with no error or warning — the coordinator reasonably believes it was saved.

arturasmckwcz (commented):
Automated code review — 5 findings (4 inline below on changed lines; 1 additional finding on unchanged code posted as a separate comment since GitHub can't anchor an inline comment outside the diff context).

arturasmckwcz:
**correctness** — `src/server/routes/agent/agent.routes.ts:283` (`GET /agent/:id`)

`GET /agent/:id` has no visibility filter for coordinator-created/unclaimed agents, unlike the new filter added to `GET /agent` (list). This line isn't part of the diff so it can't be anchored inline.

Failure scenario: any authenticated non-coordinator/admin user who knows or guesses a bare agent's sequential integer id (e.g. via the FE's `/dashboard/agents/[id]` route, which calls `GET /agent/${id}` directly) can fetch its full details even though the PR's stated goal is that such agents must stay invisible to non-coordinator/admin callers until claimed.

arturasmckwcz:
Pushed a fix for all 5 findings above (aabc8c7):

1. **joinAgent misfiring on legacy agents** — added a persisted `Agent.unclaimed` flag (migration `1787257727533-add-unclaimed-to-agent`), set only by the coordinator/admin bare-create path. `joinAgent`, `GET /agent`, and `GET /agent/:id` now gate on this flag instead of "zero `AgentPerson` rows", so legacy agents from `POST /opportunity/legacy` stay joinable.
2. **`GET /agent/:id` visibility** — now 404s an unclaimed agent for non-coordinator/admin callers, mirroring the list filter.
3. **Pagination count mismatch** — the visibility filter is now pushed into the `where` clause instead of applied in-memory after `skip`/`take`, so `count` matches the returned page.
4. **try/catch convention violation** — `POST /agent` no longer hand-builds 409/500 responses. `AgentAddressConflictError` and a new `AgentTitleConflictError` now extend `BaseError` (with a generic `details` field the global error handler spreads into the response), so they propagate naturally.
5. **`phone` silently discarded** — `createAgentBodySchema` no longer declares `phone`.

Added/updated tests in `write-agent-registration.test.ts` and `agent.routes.test.ts` covering each fix; full suite (`yarn test:run`, 702 tests) and `yarn typecheck` pass.

arturasmckwcz:
Pushed a fix for the 4 be-only findings from the follow-up review (ee73ed4):

1/2. **Existence leaks via sibling routes** — added a shared `assertAgentVisible()` helper and applied it to every `/agent/:id`-scoped route (`opportunity-linked`, `volunteer-linked`, `communication`, `contact`), not just `GET /agent/:id`. While auditing this I found the same gap in two more routes not called out explicitly (`agent-communication.routes.ts`, `agent-volunteer.routes.ts`) and fixed those too, since a partial fix would leave the "invisible until claimed" guarantee still broken.
3. **`joinAgent` 403 vs 404** — a nonexistent `agentId` now 404s instead of reusing the "not claimed yet" 403 message.
6. **`register.routes.ts` try/catch** — `createAgentForPerson` now translates a unique-title violation into `AgentTitleConflictError` internally (mirroring `createAgent`), so the route no longer hand-builds responses in a catch block. Also fixed the self-registration search picker, which had the identical "zero AgentPerson rows ≠ unclaimed" bug this PR already fixed in `joinAgent` — it was wrongly hiding legacy agents from the picker too.

Added regression tests for all of the above; full suite (707 tests) and typecheck pass.

Remaining 2 findings (exposing `unclaimed` in the agent DTOs, and a proper SDK type for `POST /agent`'s response) both require a contract change in the `sdk` repo per this org's shared rules — those need an npm publish, which I'm checking on separately before proceeding.

arturasmckwcz:
Pushed the last piece (2b71338), consuming the newly published \`need4deed-sdk@0.0.145\`:

- \`dtoAgentGetList\` now sets \`unclaimed\`, so a coordinator/admin can distinguish an unclaimed agent from a normal one via \`GET /agent\` / \`GET /agent/:id\`. Also registered it on \`sdk-types.json\`'s \`AgentList\`/\`ApiAgentGetList\`/\`ApiAgentGet\` — otherwise fast-json-stringify silently strips it from the actual response regardless of what the DTO sets (same bug class as be#575). Added a test asserting it round-trips through the real HTTP response, not just the DTO/TS layer.
- \`POST /agent\`'s success reply is now typed against the SDK's \`ApiAgentCreateResponse\` instead of being untyped.

All 6 review findings on this PR are now addressed. Full suite (707 tests) and typecheck pass.

Heads-up: while wiring this up I noticed \`ApiAgentCreateResponse\` ended up declared twice in \`sdk\`'s \`agent.ts\` — a second PR (sdk#198) added the same type in parallel with mine (sdk#202) and both merged. Harmless (TS interface merging, identical shape), but I'll clean up the duplicate in a follow-up sdk PR.

need4deed (approved):
Reviewed and fixed across two rounds — all findings addressed, full test suite passing.

PR #892, opened 2026-08-20
feat: filter opportunities by skill

### LINKED ISSUE #888: Opportunity filters overwrite each other: language, district and activity share one key
Found while building the skills filter for  issue  [905](https://github.com/need4deed-org/fe/issues/905)   pr [936](https://github.com/need4deed-org/fe/pull/936) That PR needs a `skill` branch added to`getOpportunityWhere`, and while writing it I hit a bug already on `develop`.

## The bug

Filtering opportunities by two of language, district or activity at the same time silently drops one of them. No error, just wrong results.

**Reproduce:** tick one activity and note the count, then also tick a language that does not apply to it. Expected 0, actual is the same count as the activity alone. Measured locally: activity alone 1, activity plus a non-matching language still 1.

  ## Cause

  `src/server/utils/data/get-opportunity-where.ts` builds the where-clause from
  separate spreads, and three of them write the same `deal` key:

  ```ts
  ...(filter?.language ? { deal: { dealLanguage: ... } } : {}),
  ...(filter?.district ? { deal: { dealDistrict: ... } } : {}),
  ...(filter?.activity ? { deal: { dealActivity: ... } } : {}),
 ```

  Object spread merges the top level and replaces everything below it, so when two
  are set the later one wins outright and the earlier constraint never reaches the
  query. Activity is last, so it beats both.

  Fix

One shared deal object with the constraints as optional properties inside, so they accumulate. Adding the skill branch fe#936 needs is then one more propert in the same object rather than a fourth competing spread.

I have both working locally with a unit test that fails on develop and passes with the fix. I could not open the PR: the pre-push hook runs the full suite and it does not pass on my machine. getDateObj expects day 11 and I get 10, and the onetimer test expects 2026-08-01 and I get 2026-07-31, both one day off because I
am in UTC-5 and those tests assume a European timezone. swagger.test.ts also times out at 5s intermittently. 

Blocks [936](https://github.com/need4deed-org/fe/pull/936)

ivannissimrch:
## Description
 - `filter[skill]` was already accepted by the querystring schema and the handler type, but `getOpportunityWhere` never read it, so the param was silently ignored and the filter did nothing
- Adds the `dealSkill` branch so skill actually constrains the query. This is the BE half of fe#905
 - Adding it surfaced an existing bug: language, district, and activity each wrote the same `deal` key from separate spreads, so the object spread replaced the previous constraint and only the last one reached the query. Combining an activity with a non-matching language returned the activity count instead of 0. Measured locally:
 activity alone 1, activity plus non-matching language still 1
- Fixed by giving all four constraints one shared `deal` object as optional properties, so they accumulate instead of overwriting. Skill drops in as one more property  rather than a fourth competing spread
- Unblocks fe [936](https://github.com/need4deed-org/fe/pull/936),  which needs a skill branch added to this same function. It now slots in as one more property instead of a fourth competing spread.

## Related Issues
Closes #888 

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/server/utils/data/get-opportunity-where.ts
+++ b/src/server/utils/data/get-opportunity-where.ts
@@ -24,36 +24,36 @@ export function getOpportunityWhere(
           title: ILike(`%${filter.search}%`),
         }
       : {}),
-    ...(filter?.language
+    // language, district, activity and skill all constrain the same `deal` relation, so they must share a
+    // single `deal` key.
+    // if we use two spreads with the same key, the second replaces the first, and the earlier
+    // constraint is silently dropped.
+    ...(filter?.language ||
+    filter?.district ||
+    filter?.activity ||
+    filter?.skill
       ? {
           deal: {
-            dealLanguage: {
-              language: {
-                id: normalizeStringArrayInput(filter.language),
+            ...(filter?.language && {
+              dealLanguage: {
+                language: { id: normalizeStringArrayInput(filter.language) },
               },
-            },
-          },
-        }
-      : {}),
-    ...(filter?.district
-      ? {
-          deal: {
-            dealDistrict: {
-              district: {
-                id: normalizeStringArrayInput(filter.district),
+            }),
+            ...(filter?.district && {
+              dealDistrict: {
+                district: { id: normalizeStringArrayInput(filter.district) },
               },
-            },
-          },
-        }
-      : {}),
-    ...(filter?.activity
-      ? {
-          deal: {
-            dealActivity: {
-              activity: {
-                id: normalizeStringArrayInput(filter.activity),
+            }),
+            ...(filter?.activity && {
+              dealActivity: {
+                activity: { id: normalizeStringArrayInput(filter.activity) },
               },
-            },
+            }),
+            ...(filter?.skill && {
+              dealSkill: {
+                skill: { id: normalizeStringArrayInput(filter.skill) },
+              },
+            }),
           },
         }
       : {}),
--- a/src/server/utils/data/get-opportunity-where.ts
+++ b/src/server/utils/data/get-opportunity-where.ts
@@ -101,12 +101,46 @@ function getTypeWhere(
     : {};
 }
 
+// language, district, activity and skill all constrain the same `deal`
+// relation, so they must accumulate onto one shared `deal` key rather than
+// each writing their own top-level spread — two spreads with the same key
+// have the second replace the first, silently dropping the earlier
+// constraint. Mirrors get-volunteer-where.ts's identical dealFilter.
+function getDealWhere(
+  filter: QuerystringOpportunityFiltering["filter"],
+): Record<string, unknown> {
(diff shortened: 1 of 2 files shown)


nadavosa:
Reviewed. The fix itself is solid and worth landing:

- **Real bug, correctly diagnosed**: `language`/`district`/`activity` each spread onto the same `deal` key, so combining filters silently dropped all but the last one (verified the repro described in the PR body against the pre-fix code). Merging them into one shared `deal` object so they accumulate as properties instead of competing spreads is the right fix.
- **Test coverage is good**: covers the combined-filters regression case, the `In([...])` OR-multiple-values behavior, and the single-filter-still-works case.
- `dealSkill` slots in cleanly as one more property on the same object, matching the existing `dealLanguage`/`dealDistrict`/`dealActivity` shape — consistent with the rest of the file.

**On the conflict**: this isn't a superficial one to auto-resolve. `develop` has moved since this branched — `get-opportunity-where.ts`'s `getOpportunityWhere` picked up a second `appointment` parameter and had its `type`/date-range handling extracted into `getTypeWhere`/`getAppointmentDateWhere` helpers (#893/#894), and `get-opportunity-where.test.ts` was **newly created on develop by that same PR** with entirely different test cases (type/status/search/appointment-date/excludeAccompanying) — so both branches added a new file at the same path with unrelated content, not overlapping edits to an existing one.

Resolving this needs someone to:
1. Keep develop's `getTypeWhere`/`getAppointmentDateWhere` extraction and the `appointment` param as-is.
2. Re-apply this PR's fix — the shared `deal` object with `dealLanguage`/`dealDistrict`/`dealActivity`/`dealSkill` as properties — inside that restructured function.
3. Merge the two test files rather than picking one — combine this PR's 3 new cases with develop's existing appointment-date/excludeAccompanying suite in the same `describe` block.

@ivannissimrch — since this is a substantive merge (not just adjacent lines), can you take a pass at resolving it against current `develop`?

need4deed (approved):
Reviewed independently (three finder angles + a TypeORM-source-level check on a candidate finding). The core fix is correct and well-tested — no correctness bugs in the diff. Pushed two small follow-ups on top:

- Extracted the four deal-filter branches into a `getDealWhere()` accumulator, mirroring `get-volunteer-where.ts`'s existing identical pattern — same output shape, but structurally immune to the exact key-collision bug this PR fixes (a future 5th filter can't accidentally reintroduce it).
- Added a test encoding the issue's own literal repro (activity alone, then activity + a non-matching language) as a direct regression case.

One pre-existing, unrelated finding for the record (not from this PR, not blocking): `opportunityListQuerySchema`'s `additionalProperties: false` in `querystring.ts` is mis-nested and doesn't actually restrict anything, which is why `skill`/`activity` validate at all today. Worth its own issue if not already tracked.

Thanks for the fix — this was a real, cleanly-diagnosed bug.

Issue #681, opened 2026-06-13
feat: opportunity list — date-range filter + expose appointmentDate/Time for calendar view


nadavosa:
## Context
Linked to need4deed-org/fe#646 — the calendar page needs to fetch dated, non-accompanying opportunities for a given month.

### 1. New query params on GET /opportunity

| Param | Type | Description |
|---|---|---|
| \`appointmentDateFrom\` | ISO date string | Filter opps with appointmentDate ≥ this date |
| \`appointmentDateTo\` | ISO date string | Filter opps with appointmentDate ≤ this date |
| \`hasAppointmentDate\` | boolean | When true, return only opps where appointmentDate is not null |
| \`excludeAccompanying\` | boolean | When true, exclude accompanying volunteerType |

The FE will call with the start and end of the visible calendar month, e.g.:
\`GET /opportunity?appointmentDateFrom=2026-06-01&appointmentDateTo=2026-06-30&hasAppointmentDate=true&excludeAccompanying=true\`

### 2. Include appointmentDate and appointmentTime in the list response

Currently the opportunity list DTO (\`dtoOpportunityGetList\`) may not include \`appointmentDate\` / \`appointmentTime\`. These fields must be present in the list response so the FE can pin events on the correct calendar day without fetching each opportunity individually.

Add to list response per opportunity:
- \`appointmentDate\` (ISO string or null)
- \`appointmentTime\` (string or null)
- \`agentTitle\` (the agent/center name — for the day-click panel on the FE)

### 3. SDK
Update \`ApiOpportunityGetList\` in the SDK to include the three fields above if not already present.

## Tasks
- [ ] Add \`appointmentDateFrom\`, \`appointmentDateTo\`, \`hasAppointmentDate\`, \`excludeAccompanying\` to \`opportunityListQuerySchema\`
- [ ] Apply filters in the opportunity list handler
- [ ] Expose \`appointmentDate\`, \`appointmentTime\`, \`agentTitle\` in \`dtoOpportunityGetList\`
- [ ] Update SDK type \`ApiOpportunityGetList\`

arturasmckwcz:
All three sub-issues shipped:
- need4deed-org/sdk#199 — appointmentDate/appointmentTime added to ApiOpportunityGetList, published as need4deed-sdk@0.0.143
- #890 — exposed appointmentDate/appointmentTime (+ agentTitle, already present) in the GET /opportunity list DTO (#893, merged)
[...]

PR #893, opened 2026-08-20
feat: expose appointmentDate/appointmentTime on GET /opportunity list

### LINKED ISSUE #890: feat: expose appointmentDate/appointmentTime in GET /opportunity list DTO + response schema
### Description

Sub-issue of #681 (FE calendar view, need4deed-org/fe#646). **Blocked on**
need4deed-org/sdk#199 (`ApiOpportunityGetList.appointmentDate/appointmentTime`)
being published to npm — do not stub these fields ahead of the contract.

### Proposed solution

1. Bump `need4deed-sdk` to the version publishing sdk#199.
2. `dtoOpportunityGetList` (`src/services/dto/dto-opportunity.ts`): add
   ```ts
   appointmentDate: opportunity.onetimer ? formatDate(opportunity.onetimer.date) : null,
   appointmentTime: opportunity.onetimer ? formatTime(opportunity.onetimer.date) : null,
   ```
   reusing the existing `formatDate`/`formatTime` helpers
   (`src/services/utils/index.ts`) already used for `ApiOpportunityGet.event`.
   `onetimer` is already loaded by the list route's relations — no new join
   needed.
3. **Known gotcha (see #575 for the same class of bug):** Fastify response
   serialization goes through the hand-maintained JSON-schema mirror in
   `src/server/schema/sdk-types.json`, not the SDK's TS type directly.
   `fast-json-stringify` silently drops any field not declared there. Add
   `appointmentDate`/`appointmentTime` to the `ApiOpportunityGetList`
   schema entry in that file or the new fields will be stripped from the
   wire response even though the DTO emits them.
4. Extend the DTO test (`src/test/services/dto/dto-opportunity.test.ts`)
   and the route test added in #889 to assert both fields round-trip
   through a live `fastify.inject` call (to catch exactly this
   schema-stripping class of bug, not just the DTO unit).

### Out of scope

Querystring filters (`appointmentDateFrom`/`To`, `hasAppointmentDate`,
`excludeAccompanying`) — that's #889.

arturasmckwcz:
## Summary
- Bumps `need4deed-sdk` to `0.0.143` (sdk#199, published).
- `dtoOpportunityGetList` now derives `appointmentDate`/`appointmentTime` from `Opportunity.onetimer.date` via the existing `formatDate`/`formatTime` helpers (same pattern already used for `ApiOpportunityGet.event`); `null` when there's no `onetimer`. No query change needed — the list route already loads the `onetimer` relation.
- Registers both fields on the `ApiOpportunityGetList` entry in `src/server/schema/sdk-types.json` — `fast-json-stringify` silently strips fields missing from that hand-maintained schema mirror, the same class of bug fixed in #575.

Closes #890. Part of #681 (need4deed-org/fe#646).

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint` (pre-existing errors/warnings unrelated to this change)
- [x] `yarn test:run` — 79 files / 666 tests pass
- [x] New unit tests in `dto-opportunity.test.ts` (with/without `onetimer`)
- [x] New live-server test in `opportunity.routes.test.ts` asserting the fields round-trip through `fastify.inject` (catches schema-stripping, not just DTO output)

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.142",
+    "need4deed-sdk": "0.0.143",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/src/server/schema/sdk-types.json
+++ b/src/server/schema/sdk-types.json
@@ -1006,6 +1006,12 @@
         "agentId": {
           "type": ["integer", "null"]
         },
+        "appointmentDate": {
+          "type": ["string", "null"]
+        },
+        "appointmentTime": {
+          "type": ["string", "null"]
+        },
         "volunteerNames": {
           "type": "array",
           "items": {
@@ -1029,7 +1035,9 @@
         "location",
         "accompanyingDetails",
         "agentTitle",
-        "agentId"
+        "agentId",
+        "appointmentDate",
+        "appointmentTime"
       ]
     },
     "ApiOpportunityGet": {
--- a/src/services/dto/dto-accompanying.ts
+++ b/src/services/dto/dto-accompanying.ts
@@ -2,7 +2,7 @@ import { ApiOpportunityAccompanyingDetails, OptionById } from "need4deed-sdk";
 import District from "../../data/entity/location/district.entity";
 import DealLanguage from "../../data/entity/m2m/deal-language";
 import Accompanying from "../../data/entity/opportunity/accompanying.entity";
-import { formatDate, formatTime } from "../utils";
+import { formatAppointmentDateTime } from "../utils";
 
 export function dtoOpportunityAccompanying(
   accompanying: Accompanying,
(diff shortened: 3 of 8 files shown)


arturasmckwcz:
Code review caught a critical regression in the first commit: `ApiOpportunityGet` extends `ApiOpportunityGetList`, so its `sdk-types.json` schema inherited the new `required: [...appointmentDate, appointmentTime]` — but `dtoOpportunityGet` (backing `GET /opportunity/:id`) never set them, so `fast-json-stringify` would throw on every request to that route (no test previously exercised it).

Fixed in the follow-up commit: `dtoOpportunityGet` now sets both fields too, and the onetimer→date/time formatting is extracted into a shared `formatAppointmentDateTime` helper (also now used by `dtoOpportunityAccompanying`) so the two call sites can't drift. Added regression tests for `GET /opportunity/:id` (previously untested) and DTO-level coverage for `dtoOpportunityGet`.

Not addressed here (flagged by review, considered out of scope for this PR):
- `dtoVolunteerOpportunityGetList` / the volunteer-facing opportunity list still don't expose these fields — that endpoint's response schema (`volunteer-api-opportunity.json`) is a separately hand-curated, already-narrower subset of `ApiOpportunityGetList` (no `agentTitle`, `statusMatch`, `volunteerNames`, etc. either), so this looks like deliberate scoping rather than the same drift bug. Happy to open a follow-up if you'd like the volunteer endpoint to mirror the full contract.
- No CI/generator check ties `sdk-types.json` to the SDK's actual types, so this class of bug (already fixed once in #575) can recur on the next SDK bump. Worth a dedicated tooling issue if there's appetite.

need4deed (approved):
Reviewed via /code-review — found and fixed a critical serialization crash on GET /opportunity/:id before merge (see PR discussion). Tests pass (670/670 full suite).

PR #894, opened 2026-08-20
feat: add appointment date-range + excludeAccompanying filters to GET /opportunity

### LINKED ISSUE #889: feat: add appointment date-range + excludeAccompanying filters to GET /opportunity querystring
### Description

Sub-issue of #681 (FE calendar view, need4deed-org/fe#646). No SDK dependency —
these are BE-local querystring params, not typed in the SDK (existing
`filter.*` params like `language`/`district`/`type`/`status` aren't either).
Can be implemented independently of #199 (need4deed-org/sdk#199).

The FE calendar needs to fetch dated, non-accompanying opportunities for a
given visible month, e.g.:

```
GET /opportunity?appointmentDateFrom=2026-06-01&appointmentDateTo=2026-06-30&hasAppointmentDate=true&excludeAccompanying=true
```

### Proposed solution

1. Add top-level params to `opportunityListQuerySchema`
   (`src/server/schema/querystring.ts`) and `QuerystringOpportunityList`
   (`src/server/types/endpoint-handlers.ts`):
   - `appointmentDateFrom` (ISO date string)
   - `appointmentDateTo` (ISO date string)
   - `hasAppointmentDate` (boolean)
   - `excludeAccompanying` (boolean)

   These sit alongside `sortBy`, not inside the nested `filter` object (per
   the issue's example URL).

2. Apply them in `getOpportunityWhere`
   (`src/server/utils/data/get-opportunity-where.ts`):
   - `appointmentDateFrom`/`appointmentDateTo` → range filter on
     `onetimer.date` (TypeORM `Between`/`MoreThanOrEqual`/`LessThanOrEqual`
     as needed depending on whether one or both bounds are given).
   - `hasAppointmentDate: true` → `onetimer: Not(IsNull())` (or equivalent);
     only meaningful as an inclusion filter, not exclusion when false.
   - `excludeAccompanying: true` → `type: Not(OpportunityType.ACCOMPANYING)`;
     needs to compose correctly if `filter.type` is also supplied.

3. Tests in `src/test/server/routes/opportunity.routes.test.ts` following
   the existing `sortBy=start-date` describe block's pattern (own
   Agent/Deal/Onetimer/Opportunity fixtures with a random suffix, filter
   response down to known ids, FK-safe cleanup in `afterAll`).

### Out of scope

Exposing `appointmentDate`/`appointmentTime` in the response DTO — that's
#890, blocked on need4deed-org/sdk#199.

arturasmckwcz:
## Summary
- Adds `appointmentDateFrom`, `appointmentDateTo`, `hasAppointmentDate`, `excludeAccompanying` top-level querystring params to `GET /opportunity` for the FE calendar view (fe#646), e.g. `?appointmentDateFrom=2026-06-01&appointmentDateTo=2026-06-30&hasAppointmentDate=true&excludeAccompanying=true`.
- `getOpportunityWhere` filters on `Opportunity.onetimer.date` (`Between`/`MoreThanOrEqual`/`LessThanOrEqual`, with an inclusive end-of-day bound for `...To` so a whole calendar day matches) or the plain `onetimerId` FK column for `hasAppointmentDate` alone.
- `excludeAccompanying` defers to an explicit `filter.type` when one is given (an explicit whitelist wins); otherwise excludes `ACCOMPANYING`.
- Invalid date strings throw `BadRequestError` (400) instead of silently no-op'ing.
- No SDK change needed — like the other `filter.*` params, these aren't SDK-typed.

Closes #889. Part of #681 (need4deed-org/fe#646).

### Risk called out and verified
`GET /opportunity`'s `sortBy=start-date` path excludes `"onetimer"` from its eager-loaded relations and manually joins it under its own alias (`onetimerSort`) for ordering, specifically to avoid double-hydrating it. Adding a relation-nested `where.onetimer.date` condition could in principle have conflicted with that. Verified via an integration test combining `sortBy=start-date` with the new date-range + `excludeAccompanying` filters — TypeORM auto-joins the relation-nested `where` condition independently, with no conflict.

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test:run` — 80 files / 688 tests pass
- [x] New `get-opportunity-where.test.ts` unit tests (function was previously untested), covering date-range bounds, `hasAppointmentDate`, `excludeAccompanying` precedence, and invalid-date 400
- [x] New route-level tests, including the `sortBy=start-date` combination described above

### DIFF
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -261,7 +261,7 @@ export default async function opportunityRoutes(
             ? { order: { createdAt: "ASC" } as const }
             : undefined;
 
-      const where = getOpportunityWhere(request.query.filter);
+      const where = getOpportunityWhere(request.query.filter, request.query);
 
       logger.debug(
         `GET /opportunities called. options: ${JSON.stringify({ where })}`,
--- a/src/server/schema/querystring.ts
+++ b/src/server/schema/querystring.ts
@@ -42,6 +42,13 @@ export const opportunityListQuerySchema = {
     ...sortOrderProps,
     ...sortByProps,
     ...langProp,
+    // Calendar view (be#889): filter by Opportunity.onetimer.date range,
+    // top-level (not nested in `filter`) since they're specific to this
+    // list endpoint, matching `sortBy`.
+    appointmentDateFrom: { type: "string" },
+    appointmentDateTo: { type: "string" },
+    hasAppointmentDate: { type: "boolean" },
+    excludeAccompanying: { type: "boolean" },
     filter: {
       type: "object",
       properties: {
--- a/src/server/types/endpoint-handlers.ts
+++ b/src/server/types/endpoint-handlers.ts
@@ -60,6 +60,11 @@ export type QuerystringOpportunityList = QuerystringPaginationLanguage &
     // only (not the shared QuerystringPaginationOrdering) — sorting by
     // start date only makes sense here, not for agent/volunteer/user lists.
     sortBy?: OpportunitySortField;
+    // Calendar view (be#889): filter by Opportunity.onetimer.date range.
+    appointmentDateFrom?: string;
+    appointmentDateTo?: string;
+    hasAppointmentDate?: boolean;
+    excludeAccompanying?: boolean;
   };
 
 export interface QuerystringAgentFiltering {
--- a/src/server/utils/data/get-opportunity-where.ts
+++ b/src/server/utils/data/get-opportunity-where.ts
@@ -10,30 +10,33 @@ import {
 } from "typeorm";
 import { BadRequestError } from "../../../config/error";
 import Opportunity from "../../../data/entity/opportunity/opportunity.entity";
-import { QuerystringOpportunityFiltering } from "../../types";
+import { berlinDayBoundaries } from "../../../services/jobs/german-holidays";
+import {
+  QuerystringOpportunityFiltering,
+  QuerystringOpportunityList,
+} from "../../types";
 import { normalizeStringArrayInput } from "./for-routes";
 
-export interface OpportunityAppointmentFilter {
-  appointmentDateFrom?: string;
-  appointmentDateTo?: string;
-  hasAppointmentDate?: boolean;
-  excludeAccompanying?: boolean;
-}
+export type OpportunityAppointmentFilter = Pick<
+  QuerystringOpportunityList,
+  | "appointmentDateFrom"
+  | "appointmentDateTo"
+  | "hasAppointmentDate"
+  | "excludeAccompanying"
+>;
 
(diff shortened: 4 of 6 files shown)


arturasmckwcz:
Code review caught two correctness bugs, both fixed in the follow-up commit:

1. **Timezone bug**: `appointmentDateFrom`/`appointmentDateTo` were parsed and bounded in UTC, but this is a Berlin-based product — a UTC day boundary shifts the range by Berlin's UTC offset (e.g. `appointmentDateTo=2026-06-30` would wrongly include July 1st appointments between 00:00–01:59 Berlin time, and exclude real June 1st ones in the same window). Now reuses `berlinDayBoundaries()` (already used by `scanAccompanyNotFound` for this exact `onetimer.date` column) instead of a local UTC-only `endOfDay`.
2. **`excludeAccompanying` silently ignored whenever `filter.type` was given at all** — even if that type list explicitly included `"accompanying"`, contradicting the flag's name. Now strips `ACCOMPANYING` out of an explicit type list instead of deferring to it wholesale.

Also fixed: an empty-string date param now goes through validation (and gets rejected) instead of silently bypassing the filter, and `OpportunityAppointmentFilter` is now `Pick<QuerystringOpportunityList, ...>` instead of a hand-duplicated interface that could drift.

Flagged by review but **not changed** here (pre-existing / out of scope):
- `opportunityListQuerySchema`'s `additionalProperties: false` is nested one level too deep (inside `properties`, as a sibling of `filter`, instead of alongside `properties`), so unknown querystring keys on this whole endpoint are silently accepted — confirmed pre-existing on `develop`, not introduced here. Not fixing inline since correcting it changes validation behavior for every existing param on this endpoint, not just the 4 new ones — real (if probably small) regression risk for any currently-tolerated stray param. Worth its own issue if there's appetite.
- Combining the new relation-nested `where.onetimer.date` with `sortBy=start-date` causes TypeORM to emit two separate `LEFT JOIN`s onto `onetimer` (verified via `.getSql()`) instead of reusing the sort branch's manual `onetimerSort` alias — correct (onetimer is to-one, no row duplication) but a bit wasteful. Not worth the larger route refactor to dedupe for this PR.
- Review noted the SDK-first justification in the description was imprecise: `filter.type`/`filter.status` in this same schema *are* SDK-enum-backed (`OpportunityType#`/`OpportunityStatusType#`), unlike the new scalar params — but those match the already-established untyped-scalar precedent (`search`, `district`, `agentId`, `german`), so no SDK contract change is actually needed here.

Full suite re-verified after the fix: 690/690 tests pass.

need4deed (approved):
Reviewed via /code-review — found and fixed two correctness bugs before merge (UTC vs Berlin day-boundary timezone bug, and excludeAccompanying being silently ignored when filter.type was also set). See PR discussion for details and what was deliberately left out of scope. 690/690 tests pass.

PR #896, opened 2026-08-21
Make build-bootstrap callable + publish be's develop digest

arturasmckwcz:
## Summary
- `build-bootstrap.yaml` gains `workflow_call` (existing `push`/`workflow_dispatch` triggers untouched) plus a digest output, mirroring `n4d-infra/.github/workflows/build-push-docker-image.yml`'s existing `workflow_call`/`outputs` pattern.
- `deploy-aits.yaml` now publishes the just-pushed `be:develop` digest as an org-level `BE_DEVELOP_DIGEST` variable (visible to `infra`), so the prod-release pipeline can promote whatever's currently on `develop` without scraping GHCR — which turned out to be unreliable even for an org member's PAT during today's incident (403 on package read despite a successful `docker login`).

## Requires (before this is fully live)
A new `PAT` secret in this repo's settings: a fine-grained PAT with **only** "Organization variables: read and write" — narrowest scope that lets the new step in `deploy-aits.yaml` call `gh variable set`. Until that secret exists, that one step will fail (the rest of `deploy-aits.yaml` is unaffected).

## Context
Part of automating the manual prod-release process (build fe + bootstrap → resolve be's digest → PR the digest bump into `infra/overlays/prod/kustomization.yaml`), after today's prod outage caused by stale/placeholder digests nobody had a reliable way to check before applying.

### DIFF
--- a/.github/workflows/build-bootstrap.yaml
+++ b/.github/workflows/build-bootstrap.yaml
@@ -6,6 +6,11 @@ on:
     paths:
       - data-bootstrap/**
   workflow_dispatch:
+  workflow_call:
+    outputs:
+      digest:
+        description: Digest of the pushed bootstrap image
+        value: ${{ jobs.build.outputs.digest }}
 
 jobs:
   build:
@@ -13,6 +18,8 @@ jobs:
     permissions:
       contents: read
       packages: write
+    outputs:
+      digest: ${{ steps.build.outputs.digest }}
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Same reusable-workflow treatment as fe#946, plus publishing be:develop's digest as an org variable for the prod pipeline. Confirmed the digest-publish step is continue-on-error, so deploy-aits's actual dev rollout can't be broken by the still-missing PAT secret. LGTM — merging; PAT secret still needs to be added separately for the publish step to actually take effect.

PR #898, opened 2026-08-21
Fix build-bootstrap's checkout for cross-repo workflow_call

arturasmckwcz:
## Summary
The first live test of `infra`'s new `release-prod.yaml` orchestrator failed here too: calling this workflow cross-repo via `uses: need4deed-org/be/.github/workflows/build-bootstrap.yaml@develop` checked out **infra's** repository by default (a documented `actions/checkout` gotcha for cross-repo `workflow_call`), not be's — `data-bootstrap/Dockerfile` didn't exist in what got checked out, so the build failed immediately with `lstat data-bootstrap: no such file or directory`.

Pins `actions/checkout@v4` to `repository: need4deed-org/be, ref: develop` explicitly. No behavior change for standalone `push`/`workflow_dispatch` runs.

Companion fix: fe#948 (same root cause).

### DIFF
--- a/.github/workflows/build-bootstrap.yaml
+++ b/.github/workflows/build-bootstrap.yaml
@@ -21,7 +21,14 @@ jobs:
     outputs:
       digest: ${{ steps.build.outputs.digest }}
     steps:
+      # Explicit repository+ref: a reusable workflow called cross-repo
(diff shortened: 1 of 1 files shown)


need4deed (approved):
Same fix as fe#948 for the same root cause, confirmed by the failed release-prod.yaml test run. LGTM.

PR #900, opened 2026-08-22
fix: exclude INACTIVE agents from registration picker; mask their opportunities/volunteers

### LINKED ISSUE #885: fix: exclude inactive agents from registration picker; mask their opportunities/volunteers on the profile page
## Description

Two related gaps, both scoped strictly to `AgentEngagementStatusType.INACTIVE` (not `UNRESPONSIVE` or any other status):

1. **Registration picker still shows inactive NGOs.** `GET /agent/register/search` (`src/server/routes/agent/register.routes.ts:80-113`) — the self-registration "pick your org from the list" flow used when a new user registers — queries every agent with no status filter at all:
   ```ts
   const candidates = await fastify.db.agentRepository
     .createQueryBuilder("agent")
     .leftJoinAndSelect("agent.address", "address")
     .getMany();
   ```
   `searchAgentCandidates` (`src/server/utils/data/get-agent-by-postcode.ts:96-111`) then narrows purely by street/title text match — no status check either. `fe`'s `useAgentAddressLookup.ts` renders whatever comes back as-is. Net effect: a new user can join (or be routed toward) an NGO that's tagged inactive.

2. **An inactive agent's opportunities/volunteers are still fully visible on its profile page.** `GET /agent/:id/opportunity` (`agent-opportunity.routes.ts`) and `GET /agent/:id/volunteer` (`agent-volunteer.routes.ts`) return full data with no consideration of the agent's `engagementStatus`.

## Why

Once an NGO is marked inactive, it shouldn't be offered to new registrants, and its opportunities/volunteers shouldn't be presented as live, actionable data on its profile page.

## Requirements

- Exclude agents with `engagementStatus === AgentEngagementStatusType.INACTIVE` from `GET /agent/register/search`'s candidates.
- On `GET /agent/:id/opportunity` and `GET /agent/:id/volunteer`, when the agent is `INACTIVE`: don't drop the records, but mask their identifying fields the same way existing PII masking already does — reuse `maskString()`/`maskFields()` from `src/server/utils/pii/mask.ts` (single random char + `***`) rather than inventing a new masking scheme.
- **Must be a live check against the agent's current `engagementStatus`, not a one-time/frozen decision** — no migration, no snapshotting. The moment a coordinator flips the status back to `ACTIVE`, both behaviors must reverse immediately: the agent reappears in the registration picker, and its opportunities/volunteers show unmasked again on its next load.
- Scope is strictly `INACTIVE` — `UNRESPONSIVE` and other statuses are unaffected by this issue.

## Checklist
- [ ] Filter `INACTIVE` agents out of `GET /agent/register/search`
- [ ] Mask opportunity/volunteer identifying fields on `GET /agent/:id/opportunity` and `GET /agent/:id/volunteer` when the agent is `INACTIVE`, reusing `mask.ts`'s existing primitives
- [ ] Confirm both behaviors are driven live off current `engagementStatus` (verify a status flip back to `ACTIVE` immediately restores both)

arturasmckwcz:
## Description
Two related gaps, both scoped to `AgentEngagementStatusType.INACTIVE`:

- `GET /agent/register/search` queried every agent with no status filter — an INACTIVE agent was fully offered to new registrants. Now excluded alongside the existing `unclaimed` check.
- `GET /agent/:id/opportunity-linked` and `GET /agent/:id/volunteer-linked` returned full, unmasked data regardless of the agent's status. Now mask the opportunity's `title` and each linked volunteer's `name`/`avatarUrl` (via `maskFields()`/`maskString()` from `mask.ts` — no new masking scheme) for non-coordinator/admin callers when the agent is INACTIVE.

Both checks read live off the agent's current `engagementStatus` — no snapshotting. Flipping the status back to `ACTIVE` immediately reverses both behaviors (covered by tests).

`maskFields()` and `PERSON_PII_FIELDS` are now exported from `mask.ts` so the two routes can reuse them.

## Related Issues
Closes #885

## Checklist
- [x] Within scope of the issue; no unrelated files touched
- [x] Tests added (new `agent-opportunity.routes.test.ts`, new `agent-register-search.routes.test.ts`, extended `agent-volunteer.routes.test.ts`) covering: ACTIVE unmasked, INACTIVE masked for a non-coordinator, INACTIVE unmasked for coordinator/admin, and re-activation immediately restoring visibility
- [x] `yarn typecheck` / `yarn lint` clean on touched files
- [x] Full suite green: 83 files / 713 tests passing locally against Postgres


need4deed (approved):
Approved — two independent review rounds completed, all findings addressed (null-guard fix, plus two INACTIVE-agent bypass fixes: joinAgent() and GET /opportunity/:id/volunteer-linked). Full suite green (85 files / 720 tests), typecheck/lint clean.

PR #907, opened 2026-08-22
feat: GET /event — public list of events (be#903)

arturasmckwcz:
## Summary
- `GET /event`, public — bumps `need4deed-sdk` 0.0.145 → 0.0.146 to consume `ApiEventN4DGetList`.
- New `tryAuthenticate()` decorator on the jwt plugin: unlike `authenticate()`, it best-effort resolves `request.authUser` from a cookie if present but never requires/throws on one. This is what lets this single public route vary its response for a logged-in coordinator — anonymous/non-privileged callers see only `isActive` events, coordinator/admin see everything (including drafts/deactivated events, since the dashboard needs to manage those too).
- Translation resolution: `?language=<lang>` (defaults to `Lang.DE`) resolves each event's matching `EventTranslation` row; falls back to whatever translation exists rather than dropping the event from the feed if it hasn't been authored in the requested language yet. An event with zero translations is excluded entirely.
- Response uses the standard `{message, data, count}` envelope (`ReplyDataCount<ApiEventN4DGetList[]>`), matching every other list route in this repo.
- New `event.schema.ts` — inline, not `$ref`'d into `sdk-types.json` (per discussion: that file's already too cluttered for a brand-new domain).
- New `QuerystringEventGetList` type (`endpoint-handlers.ts`) — `QuerystringPaginationLanguage` doesn't fit since this route has no pagination/ordering and `language` is optional.
- `RoutePrefix.EVENT = "/event"` — matches `website`'s `urlApiEvent`, not the `/event-n4d` this issue originally named.

## Related
Part of epic #458.

## Checklist
- [x] `yarn typecheck` / `yarn lint` clean on touched files
- [x] Tests added (`event.routes.test.ts`): standard envelope shape (including count), active-only filtering for anonymous + non-privileged callers, full list for coordinator, language resolution (`de` default, `en` requested), fallback when the requested language isn't authored, exclusion of an event with zero translations
- [x] Full suite green: 86 files / 730 tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "class-validator": "^0.14.2",
     "fastify": "^5.3.3",
     "fastify-plugin": "^5.0.1",
-    "need4deed-sdk": "0.0.145",
+    "need4deed-sdk": "0.0.146",
     "node-cron": "^4.5.0",
     "nodemailer": "^9.0.3",
     "pg": "^8.14.1",
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -20,6 +20,7 @@ import appreciationRoutes from "./routes/appreciation.routes";
 import authRoutes from "./routes/auth";
 import commentRoutes from "./routes/comment";
 import communicationRoutes from "./routes/communication.routes";
+import eventRoutes from "./routes/event.routes";
 import healthRoutes from "./routes/health";
 import activityLogCollectionRoutes from "./routes/m2m/activity-log.routes";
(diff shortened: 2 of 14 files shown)


need4deed (approved):
Approved — independent review found and fixed a real security gap (tryAuthenticate() didn't check the JWT type claim, so a leaked never-expiring email-verification token could grant permanent privileged access) and a real correctness bug (coordinator couldn't see untranslated/draft events, contradicting the PR's own stated intent). Follow-up for the wider authenticate() gap backlogged as #908. Full suite green (86 files / 731 tests).

PR #910, opened 2026-08-23
feat: POST /event — coordinator create (be#904)

arturasmckwcz:
## Summary
- \`POST /event\`, \`fastify.authenticate({ role: UserRole.COORDINATOR })\`.
- Body: \`ApiEventN4DCreate\` (\`need4deed-sdk@0.0.146\`) — structural fields plus \`translations: ApiEventN4DTranslationInput[]\` (at least one, enforced via the request schema's \`minItems\` rather than application code).
- New \`write-event.ts\`: \`createEvent()\` writes one \`EventN4D\` row + one \`EventTranslation\` row per submitted language in a single transaction.
- **Naming deviation from the issue text**: this issue originally suggested \`parser-event.ts\`, but since the function actually performs the DB write (not just an input→entity transform), I put it in \`src/server/utils/data/write-event.ts\` instead, matching \`write-agent-registration.ts\`'s existing convention (\`write-*.ts\` = does I/O, \`parser-*.ts\` under \`services/dto/\` = pure transform).
- \`active\` defaults to \`false\` (matching \`EventN4D.isActive\`'s own column default), not \`true\` — a created event starts as a draft unless the coordinator explicitly publishes it, consistent with why a coordinator can see inactive events at all (be#903).
- Response reuses GET /event's \`dtoEventN4DGetList\`/\`eventItemSchema\`, echoing back in whichever language was submitted first.
- New \`eventCreateBodySchema\`/\`eventCreateResponseSchema\` (\`event.schema.ts\`), inline like the rest of this file, per the discussion on #907.

## Related
Part of epic #458.

## Checklist
- [x] Tests added (\`event-create.routes.test.ts\`): single-language create (defaults inactive), multi-language create (respects explicit \`active: true\`), rejects non-coordinator, rejects empty \`translations\`
- [x] Full suite green: 87 files / 735 tests

### DIFF
--- a/src/server/routes/event.routes.ts
+++ b/src/server/routes/event.routes.ts
@@ -1,9 +1,20 @@
 import { FastifyInstance, FastifyPluginOptions } from "fastify";
-import { ApiEventN4DGetList, Lang, UserRole } from "need4deed-sdk";
+import {
+  ApiEventN4DCreate,
+  ApiEventN4DGetList,
+  Lang,
+  UserRole,
+} from "need4deed-sdk";
+import { NotFoundError } from "../../config";
 import { dtoEventN4DGetList } from "../../services";
-import { eventListResponseSchema, langQuerySchema } from "../schema";
-import { QuerystringEventGetList, ReplyDataCount } from "../types";
-import { getLanguageCode } from "../utils";
+import {
+  eventCreateBodySchema,
+  eventCreateResponseSchema,
+  eventListResponseSchema,
+  langQuerySchema,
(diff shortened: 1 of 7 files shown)


need4deed (approved):
Approved — independent review found and fixed a real completeness bug (create response was silently dropping 6 accepted+persisted fields), an error-handling gap (raw TypeORM error instead of a BaseError subclass), a data-integrity gap (duplicate-language translations), and a missing date-range validation. Full suite green (87 files / 737 tests).

PR #911, opened 2026-08-23
feat: PATCH /event/:id — coordinator update (be#905)

arturasmckwcz:
## Summary
- \`PATCH /event/:id\`, \`fastify.authenticate({ role: UserRole.COORDINATOR })\`, \`204 No Content\` on success — matches this repo's dominant PATCH convention (agent, opportunity, organization all do the same).
- Body: \`ApiEventN4DPatch\` (\`need4deed-sdk@0.0.146\`). Structural fields are a plain partial update: omitted = unchanged, explicit \`null\` clears a nullable column. \`date\`/\`type\`/\`linkRSVP\`/\`address\`/\`active\` reject \`null\` in the schema even though the SDK's \`VoidableProps\` type technically allows it — those are required columns, and "clear this required field" isn't a meaningful request.
- \`translations\` is an upsert per \`(event, language)\`: an included entry replaces that language's \`EventTranslation\` row entirely if one exists, otherwise inserts a new one; languages not mentioned are left untouched — per the issue's stated design.
- Extracted \`translationFields\`/\`assertDistinctLanguages\` in \`write-event.ts\`, shared with \`createEvent\` (be#904) rather than duplicated.
- \`dateEnd\`/\`date\` cross-validation accounts for partial updates — compares the *effective* date/dateEnd (patch value if given, otherwise the existing row's), since either one may be omitted from the request.

## Related
Part of epic #458.

## Checklist
- [x] Tests added (\`event-patch.routes.test.ts\`): structural-only patch leaves translations untouched, updates an existing language in place, adds a new language without touching others, clears a nullable field via explicit null, rejects non-coordinator, 404s a nonexistent id, rejects duplicate languages, rejects an invalid date range
- [x] Full suite green: 88 files / 745 tests

### DIFF
--- a/src/data/entity/event/event_translation.entity.ts
+++ b/src/data/entity/event/event_translation.entity.ts
@@ -4,11 +4,16 @@ import {
   JoinColumn,
   ManyToOne,
   PrimaryGeneratedColumn,
+  Unique,
 } from "typeorm";
 import Language from "../profile/language.entity";
 import EventN4D from "./event.entity";
 
+// One translation per (event, language) — be#905 review found the upsert in
+// write-event.ts was a find-then-insert with no DB-level guard, so a
+// concurrent PATCH race could silently create duplicates.
 @Entity()
+@Unique(["eventn4dId", "languageId"])
 export default class EventTranslation {
   constructor(event?: Partial<EventTranslation>) {
     if (event) {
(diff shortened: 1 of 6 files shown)


need4deed (approved):
Approved — independent review found and fixed two real high-severity bugs (a stale-translation-field bug and a concurrent-write data-integrity race, closed with a new DB unique constraint + migration) plus two consistency/efficiency gaps. Full suite green (88 files / 747 tests).

PR #912, opened 2026-08-24
feat: DELETE /event/:id — coordinator delete (be#906)

arturasmckwcz:
## Summary
- \`DELETE /event/:id\`, \`fastify.authenticate({ role: UserRole.COORDINATOR })\`.
- **Design decision (confirmed before implementing, per the issue's own flag): hard delete, not soft-deactivate.** \`PATCH /event/:id { active: false }\` (be#905) already covers "hide without losing data" — a soft-delete DELETE would just be a second way to do the same thing. This is a real row delete, cascading to \`EventTranslation\` via the entity's existing \`onDelete: "CASCADE"\`.
- Mirrors \`DELETE /agent/:id\`'s exact shape — inline in the route (no separate \`write-event.ts\` helper needed for a plain existence-check + delete) and its \`200 + { message }\` response convention (this repo's DELETE and PATCH conventions differ: 200/message vs 204/no-body — this follows DELETE's).

## Related
Closes epic #458 — this is the last of its 4 sub-issues.

## Checklist
- [x] Tests added (\`event-delete.routes.test.ts\`): deletes the row and cascades to translations, rejects non-coordinator, 404s a nonexistent id
- [x] Full suite green: 89 files / 750 tests

### DIFF
--- a/src/server/routes/event.routes.ts
+++ b/src/server/routes/event.routes.ts
@@ -23,6 +23,7 @@ import {
   QuerystringEventGetList,
   ReplyData,
   ReplyDataCount,
+  ReplyMessage,
 } from "../types";
 import { createEvent, getLanguageCode, updateEvent } from "../utils";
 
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Approved — independent review found and fixed two real test-quality gaps (unasserted fixture failure, redundant cleanup), declined a systemic TOCTOU pattern shared with the pre-existing DELETE /agent/:id with reasoning, and caught a missing-assignee process gap fixed across several PRs/issues this session. Full suite green (89 files / 750 tests).

PR #916, opened 2026-08-25
feat: scope GET /opportunity to an AGENT caller's own agents

ivannissimrch:
## Description
fe [934](https://github.com/need4deed-org/fe/issues/934) asks that NGO users only see opportunities belonging to their own organisation. Today `GET /opportunity` applies no caller-based scoping at all, so an NGO user sees every organisation's.

The issue says both "their own operator (their email domain)" and "for their own shelter", which give different lists. Nadav settled it on fe [950](https://github.com/need4deed-org/fe/pull/950): "anyone from the same NGO/agent can see the opportunities there of that NGO/agent". So this scopes by agent. `agent.organization_id` is NULL for every agent on local and dev so that operator scoping would return an empty list for everyone today anyway.

`GET /opportunity` now resolves the caller's agent memberships server-side and restricts the result to those agents when the caller's role is AGENT. Coordinators and admins are unaffected.

Only ACTIVE memberships count. A PENDING membership is one a coordinator has not approved yet, so it grants nothing.

fe [950](https://github.com/need4deed-org/fe/pull/950) is the frontend half and needs this to be safe to ship.

## Related Issues
fe [934](https://github.com/need4deed-org/fe/issues/934)
 fe [950](https://github.com/need4deed-org/fe/pull/950)

## Changes
- `getCallerAgentIds` in `src/server/utils/data/`: a person's ACTIVE agent memberships, deduped
- `GET /opportunity`: when the caller is an AGENT, constrain `where.agent.id`
    to those IDs
- An AGENT with no ACTIVE membership gets an empty list rather than falling
    through to an unfiltered query, so the failure mode is closed, not open
- 6 tests in `src/test/server/routes/opportunity-agent-scope.routes.test.ts`

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -264,6 +264,8 @@ export default async function opportunityRoutes(
             : undefined;
 
       const where = getOpportunityWhere(request.query.filter, request.query);
+      const message = `Opportunities page:${request.query.page}.`;
+
       //  NGOs see only their own agent's opportunities
       if (request.authUser?.role === UserRole.AGENT) {
         const agentIds = await getCallerAgentIds(
@@ -275,7 +277,7 @@ export default async function opportunityRoutes(
         // skipping the filter, which would show everything.
         if (agentIds.length === 0) {
           return reply.status(200).send({
-            message: `Opportunities page:${request.query.page}.`,
+            message,
             data: [],
             count: 0,
           });
@@ -373,7 +375,7 @@ export default async function opportunityRoutes(
 
       // DTO (dtoOpportunityGetList) runs in the preSerialization hook after PII masking.
       return reply.status(200).send({
-        message: `Opportunities page:${request.query.page}.`,
+        message,
         data: opportunitiesCategoryDistrict,
         count,
       });
--- a/src/server/routes/opportunity/opportunity.routes.ts
+++ b/src/server/routes/opportunity/opportunity.routes.ts
@@ -13,6 +13,7 @@ import {
   SortOrder,
   UserRole,
 } from "need4deed-sdk";
+import { FindOptionsWhere, In } from "typeorm";
 import {
   BadRequestError,
   NotFoundError,
@@ -61,6 +62,7 @@ import {
 import {
   addAgentTypeServiceTranslations,
   addComments2Entity,
+  getCallerAgentIds,
   getCategoryToDealHandler,
   getDistrictToAgentHandler,
   getDistrictToOpportunityHandler,
@@ -262,6 +264,28 @@ export default async function opportunityRoutes(
             : undefined;
 
       const where = getOpportunityWhere(request.query.filter, request.query);
+      //  NGOs see only their own agent's opportunities
+      if (request.authUser?.role === UserRole.AGENT) {
+        const agentIds = await getCallerAgentIds(
+          fastify,
+          request.authUser.personId,
+        );
+
+        // An agent with no shelter must see nothing, so return here rather than
+        // skipping the filter, which would show everything.
+        if (agentIds.length === 0) {
+          return reply.status(200).send({
+            message: `Opportunities page:${request.query.page}.`,
+            data: [],
+            count: 0,
+          });
+        }
+        // Spread so this composes if getOpportunityWhere ever sets an agent filter;
+        where.agent = {
+          ...(where.agent as FindOptionsWhere<Agent> | undefined),
+          id: In(agentIds),
+        };
+      }
 
       logger.debug(
         `GET /opportunities called. options: ${JSON.stringify({ where })}`,
--- a/src/server/utils/data/get-caller-agent-ids.ts
+++ b/src/server/utils/data/get-caller-agent-ids.ts
@@ -1,0 +1,17 @@
+import { FastifyInstance } from "fastify";
+import { AgentMembershipStatus } from "need4deed-sdk";
+
+export async function getCallerAgentIds(
+  fastify: FastifyInstance,
+  personId: number | null | undefined,
+): Promise<number[]> {
+  if (!personId) {
+    return [];
+  }
+
+  const memberships = await fastify.db.agentPersonRepository.find({
+    where: { personId, status: AgentMembershipStatus.ACTIVE },
+  });
+
+  return [...new Set(memberships.map((m) => m.agentId))];
+}
--- a/src/server/utils/data/index.ts
+++ b/src/server/utils/data/index.ts
@@ -8,6 +8,7 @@ export * from "./create-agent-contact";
 export * from "./for-routes";
 export * from "./get-agent-by-postcode";
 export * from "./get-agent-where";
+export * from "./get-caller-agent-ids";
 export * from "./get-language-title";
 export * from "./get-opportunity-orphanage-agent";
 export * from "./get-opportunity-where";
--- a/src/test/server/routes/opportunity-agent-scope.routes.test.ts
+++ b/src/test/server/routes/opportunity-agent-scope.routes.test.ts
@@ -1,0 +1,176 @@
+import { FastifyInstance } from "fastify";
+import { AgentMembershipStatus, AgentRoleType, UserRole } from "need4deed-sdk";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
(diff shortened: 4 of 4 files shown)


src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -439,6 +471,7 @@ export default async function opportunityRoutes(
           ? await fastify.db.agentPersonRepository.findOneBy({
               agentId,
               personId,
+              status: AgentMembershipStatus.ACTIVE,
```
arturasmckwcz: **correctness (auth bypass):** good that this adds an ACTIVE-only status filter, but the `personId` feeding this check (a couple lines above, not shown in this hunk) is `request.authUser?.personId || request.body.submitted_by_id` — if the caller's own `personId` is null/undefined, it silently falls back to the client-supplied `submitted_by_id`.

Failure scenario: an AGENT-role account with no linked personId sends `agent_id: <targetAgent>` and `submitted_by_id: <activeMemberPersonId>` (a personId it knows is an ACTIVE member of the target agent); this ACTIVE-membership check then succeeds using the impersonated id, letting a non-member create an opportunity for an agent they don't belong to.

src/server/routes/opportunity/opportunity.routes.ts
```diff
@@ -725,6 +759,7 @@ export default async function opportunityRoutes(
           await fastify.db.agentPersonRepository.findOneBy({
             agentId: effectiveAgentId,
             personId: contactLinkId,
+            status: AgentMembershipStatus.ACTIVE,
```
arturasmckwcz: **correctness / regression:** the `contactLinkId` check now requires `status: AgentMembershipStatus.ACTIVE`, newly rejecting a PENDING agent member who previously could be set as an opportunity's contact — no test in the diff covers this behavior change, and the resulting error message ("is not registered as a contact") is misleading since the person IS registered, just not yet ACTIVE.

Failure scenario: a coordinator or agent tries to set a freshly-invited, still-PENDING colleague as an opportunity's contact during onboarding; the PATCH now 404s implying no registration exists at all.

src/server/routes/opportunity/opportunity-volunteer.routes.ts
```diff
@@ -36,6 +38,22 @@ export default function opportunityOpportunityVolunteerRoutes(
         throw new BadRequestError(msg400);
       }
 
+      if (request.authUser?.role === UserRole.AGENT) {
+        const opportunity = await fastify.db.opportunityRepository.findOne({
```
arturasmckwcz: **efficiency:** this new AGENT ownership check issues its own `opportunityRepository.findOne({ id, select: agentId })` immediately before `opportunityVolunteerRepository.find(...)` below, which already joins `opportunity.agent` via `relations` — an unnecessary extra DB round trip on every AGENT-role call.

(Note: a naive fix of checking `volunteers[0]?.opportunity?.agentId` instead would be unsafe on its own — when zero volunteers are linked, `volunteers[0]` is undefined regardless of ownership, silently turning the intended 404 into a 200 with an empty array.)

src/server/routes/opportunity/opportunity-volunteer.routes.ts
```diff
@@ -36,6 +38,22 @@ export default function opportunityOpportunityVolunteerRoutes(
         throw new BadRequestError(msg400);
       }
 
+      if (request.authUser?.role === UserRole.AGENT) {
```
arturasmckwcz: **simplification:** this ownership-check-and-404 block (resolve caller's agentIds, compare to `opportunity.agentId`, throw `NotFoundError`) is copy-pasted near-verbatim across three handlers in this PR (`opportunity.routes.ts` GET /:id, this file, and a list-specific variant), rather than factored into one shared helper. A future change to this security-critical rule would need to be applied by hand in three places — missing one silently reopens the cross-agent exposure this PR is meant to close.

src/server/utils/pii/visible-persons.ts
```diff
@@ -65,8 +66,9 @@ export async function resolveCallerVisibility(
   }
 
   const agentPersonRepository = fastify.db.agentPersonRepository;
-  const memberships = await agentPersonRepository.find({ where: { personId } });
-  memberships.forEach((m) => agentIds.add(m.agentId));
+  const callerAgentIds = await getCallerAgentIds(fastify, personId);
```
arturasmckwcz: **efficiency:** for AGENT-role callers, `getCallerAgentIds` now runs twice per request on `GET /opportunity/:id`, `GET /opportunity`, and `GET /opportunity/:id/volunteer-linked` — once in the route handler for the new ownership/scoping check, and again here inside `resolveCallerVisibility`'s PII masking hook wired on the same routes. Consider resolving the caller's agentIds once per request and threading it through.

Also: a few lines below (unchanged by this diff, ~line 77), `members = await agentPersonRepository.find({ where: { agentId: In([...agentIds]) } })` — which determines whose PII becomes unmasked for the caller — still has no `status` filter, contradicting the comment on `getCallerAgentIds` that a PENDING membership "grants nothing." An ACTIVE caller at Agent X with a PENDING co-applicant Y will still get Y's `personId` added to the unmasked set.

arturasmckwcz (commented):
Automated review findings (Claude Code `/code-review`). This PR tightens agent/cross-org scoping on opportunities, but the scoping is inconsistently applied — a few gaps re-open the exposure it's meant to close. Inline comments cover issues on lines this PR touches; the following pre-existing code paths are **not touched by this diff** but are directly relevant to the scoping goal and should be checked in a follow-up:

1. **`src/server/routes/comment.ts:86` — `GET /comment` has no ownership/agent-scoping check at all.** Any authenticated user (including an AGENT from an unrelated organisation) can read another agent's opportunity comment thread by numeric `entityId`. An AGENT-role caller who this PR now correctly 404s on `GET /opportunity/<otherAgentOppId>` can still call `GET /comment?entityType=opportunity&entityId=<otherAgentOppId>` and get the full comment thread (internal notes, potentially PII) for that opportunity — defeating the isolation this PR establishes elsewhere.

2. **`src/server/routes/volunteer/volunteer-opportunity.routes.ts:54` — `GET /volunteer/opportunity` has no role restriction or agent scoping.** It's registered under a parent hook that only requires `fastify.authenticate()`, and builds its `where` purely from client query filters, unlike the now-fixed `GET /opportunity`. An AGENT-role caller can hit this endpoint instead and get an unscoped, cross-organisation opportunity list.

3. **`src/server/routes/opportunity/opportunity.routes.ts:522` — `contactPersonId` freezing (`agent.agentPerson?.some(ap => ap.personId === opportunity.submittedByPersonId)`) uses a plain relation with no `status` filter**, inconsistent with the ACTIVE-only authorization check added in this PR just above it (line 474). A COORDINATOR (who bypasses the AGENT-only membership gate) creating an opportunity with `submitted_by_id` set to a person who only has a PENDING membership in the target agent will permanently freeze that not-yet-approved person in as the opportunity's public-facing contact.

src/server/utils/data/get-caller-agent-ids.ts
```diff
@@ -0,0 +1,19 @@
...
+//  one is still waiting on a coordinator to approve it, so it grants nothing.
+export async function getCallerAgentIds(
+  fastify: FastifyInstance,
+  personId: number | null | undefined,
+): Promise<number[]> {
+  if (personId === null || personId === undefined) {
+    return [];
+  }
+
+  const memberships = await fastify.db.agentPersonRepository.find({
```
arturasmckwcz: **reuse:** this re-implements a query that the existing `getActiveAgentMemberships(personId)` helper (`src/server/utils/data/get-agent-memberships.ts`) already performs verbatim (same table, same personId+ACTIVE where-clause) instead of composing on top of it — per the shared-rules "reuse before you create" convention. If the definition of "active membership" ever changes, this duplicate can silently drift out of sync.

need4deed:
@ivannissimrch can we merge this?

ivannissimrch:
@need4deed Most of the review is fixed and pushed. Two comments I haven't addressed yet: getCallerAgentIds resolving twice per request, and the legacy public route. I need help with those two; I'm not confident enough to finish them on my own.

need4deed:
@arturasmckwcz can you help   out in guiding here?

